PR #4, opened 2014-09-02
Fixing build on windows with Visual Studio

ENikS:
Building project with Microsoft tools choke on this include file:<unistd.h> 
It should be excluded when compiled with Visual Studio compiler on MS OS.

### DIFF
--- a/db/c.cc
+++ b/db/c.cc
@@ -5,7 +5,9 @@
 #include "leveldb/c.h"
 
 #include <stdlib.h>
+#ifndef _MSC_VER
 #include <unistd.h>
+#endif
 #include "leveldb/cache.h"
 #include "leveldb/comparator.h"
 #include "leveldb/db.h"


chfast:
Where can I find any instructions how to build leveldb with VS and/or MinGW?

maxd:
@chfast This is not an easy task. Here you can find bash [script](https://github.com/protonail/leveldb-jna/blob/master/bin/build_leveldb.sh) and [instruction](https://github.com/protonail/leveldb-jna/blob/master/BUILD.md) how to build LevelDB (use env_win.cc from bitcoin project) use MinGW.

ralphtheninja:
@chfast You can also have a look at `leveldown` where we do build for windows. env_win.cc is here https://github.com/Level/leveldown/blob/master/deps/leveldb/port-libuv/env_win.cc

chfast:
@ralphtheninja thanks for that, might be useful for me as I'm trying to get Python package working on Windows.

Nevertheless, is leveldb interested in having basic Windows support?

ENikS:
@chfast You could try to clone  ENikS/leveldb or ENikS/bitcoin-dev-msvc to see how it is done in MS VS. It is not up to date but will give you enough to start

ENikS:
I signed it!

pwnall:
The current code builds on Windows, as tracked by AppVeyor. Going to close this. Sorry we weren't able to get to it earlier!

PR #275, opened 2015-01-11
readme: improved documentation link

paulirish:
This replaces htmlpreview with [rawgit](http://rawgit.com/).
rawgit is faster and doesnt use JS to render the page, making it SEO friendly.


paulirish:
ping @cmumford. 

PTAL?

paulirish:
:tada:

PR #277, opened 2015-01-12
allow flags from environment

stevejames:
Allow flags from the environment


cmumford:
@stevejames:

If I'm reading the [override docs](https://www.gnu.org/software/make/manual/html_node/Override-Directive.html) correctly this change does the opposite of what the description says.

pwnall:
This PR references the now-removed Make build system. Going to close it.

PR #278, opened 2015-01-29
block_builder.cc header file dependency fixed

wankai:


cmumford:
Hi @wankai:

1. Need to have a space after main include file (table/block_builder.h), and also...
2. Headers need to be an alphabetical order.

cmumford:
@googlebot rescan

PR #284, opened 2015-04-20
log compaction output file's level along with number

ideawu:
The `FinishCompactionOutputFile()` method write a log like this:

```
Generated table #132: 803288 keys, 21119441 bytes
```
The level of the output file is missing, I change it to write log message like this:

```
Generated table #132@1: 803288 keys, 21119441 bytes
```

Which is similar to other logs, like

```
Compacted 1@1 + 1@2 files => 17115883 bytes
```


ideawu:
I signed it!

ralphtheninja:
@cmumford Time for 1.19 soonish?

bwzhang2011:
@cmumford, hope 1.19 could be provided since 1.18 released 2 years ago, would you mind leaving us such time schedule ?

PR #286, opened 2015-04-26
Android NDK port for LevelDB.

whip112:
gather related code run yahpay-fastfix and rebuild using whip112 - build tools


whip112:
build project

PR #296, opened 2015-06-05
BuildBatchGroup no longer skip NULL batch, for TEST_CompactMemTable

bluebore:


cmumford:
@googlebot rescan

bluebore:
I signed it!

cmumford:
@bluebore I don't see how your change "no longer skip[s] NULL batch". In fact it will break out of the for-loop early on the first null write batch. Do you have a test (or program) that fails without this change?

bluebore:
@cmumford  In "DBImpl::TEST_CompactMemTable" , we use a "null write batch" to ensure that "DBImpl::Write" calls "DBImpl::MakeRoomForWrite" with the "true" parameter. "null write batch" should be handled by "DBImpl::Write" instead of "DBImpl::BuildBatchGroup". So breaking out of the for-loop early on the first “null write batch” is what we expected.

PR #300, opened 2015-08-02
CMake build support.Windows support.

rajasrijan:
CMake build support.Windows support.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,78 @@
+cmake_minimum_required (VERSION 2.8)
+project (leveldb)
+
+file(GLOB SOURCES1 "db/*.cc")
+file(GLOB TESTS db "db/*test*.cc")
+file(GLOB BENCH db "db/*_bench.cc")
+list(REMOVE_ITEM SOURCES1 ${TESTS} ";" ${BENCH})
+
+file(GLOB SOURCES2 "util/*.cc")
+file(GLOB TESTS db "util/*test*.cc")
+file(GLOB BENCH db "util/*_bench.cc")
+file(GLOB POSIX db "util/*_posix.cc")
+list(REMOVE_ITEM SOURCES2 ${TESTS} ";" ${BENCH})
+
+file(GLOB SOURCES3 "table/*.cc")
+file(GLOB TESTS db "table/*test*.cc")
+file(GLOB BENCH db "table/*_bench.cc")
+list(REMOVE_ITEM SOURCES3 ${TESTS} ";" ${BENCH})
+
+if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
+	# using Visual Studio C++
+	list(APPEND SOURCES3 "port/port_win.cc")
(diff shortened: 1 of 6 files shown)


rajasrijan:
I signed it!

cmumford:
Hi @rajasrijan:

Thanks for submitting this pull request. For the time being the leveldb project intends to maintain support only for POSIX (on Linux). Please see [Contributing to the leveldb Project](https://github.com/google/leveldb#contributing-to-the-leveldb-project) for more details.

Qix-:
Erm, sorry to be 'that guy' but isn't this a win for the adoption of LevelDB? Otherwise, I'm forced to go with some other library for cross-platform support. I understand the desire to keep things 100% POSIX - that's great! - but unfortunately that's not _absolutely_ possible with Windows.

This PR seems pretty acute and unobtrusive and the adoption of CMake is quite modern as Make is (very slowly) being dropped due to compatibility concerns.

If this is all it takes to make LevelDB support Windows, then why not?

juliangruber:
@Qix- if you don't agree with the maintainers' decisions - which can be manifold - this is still open source, so you're free to maintain your own fork with Windows support. And others will be thankful for your actions too! Also, I'm quite sure there's a working Windows fork somewhere already.

LevelDB _primarily_ is made for Chrome, and hence there doesn't need to be Windows support as Chrome provides it's own env for LevelDB. If you imagine now @cmumford adds Windows support by popular request, he'd also have to maintain it if he wants to keep the quality of this project high. And as that's not relevant to Chrome it would be hard to justify - and very likely not interesting at all.

Qix-:
That's not the point, though. If this was just for Chrome, and that's the justification for cutting out an entire first-world operating system, then it should be included in the readme (which currently doesn't mention any adherence to the Chrome environment).

As well, it's perfectly fine to question a maintainer's decisions - oftentimes a maintainer doesn't understand the need for a particular feature, or might be interested in entertaining the idea of making a tradeoff. In this case, the entirety of the library would still be purely POSIX except in the case of Windows, which would be the case to begin with.

In return, this opens up adoption on Windows for applications that aren't running in a Chrome environment (i.e. anything other than Chrome).

As demonstrated in this PR, the additional code necessary for Windows support is minimal. I think you're overestimating the amount of maintenance that would be required for this - and _under_estimating the willingness of the open source community, _especially_ with a Google library.

juliangruber:
When it gets public that LevelDB supports Windows, people will run it on Windows and will very likely hit bugs that only appear in the Windows environment. When then there is no maintainer interested in debugging those issues, the project's quality will decline by the number of open requests.

> As well, it's perfectly fine to question a maintainer's decisions - oftentimes a maintainer doesn't understand the need for a particular feature, or might be interested in entertaining the idea of making a tradeoff. In this case, the entirety of the library would still be purely POSIX except in the case of Windows, which would be the case to begin with.

Open source is no democracy, it's 100% up to the maintainer to make whatever decisions they decide to be appropriate. There's no obligation for the maintainer to do otherwise. There's no warranty attached, see also the license: https://github.com/google/leveldb/blob/master/LICENSE#L18-L20

PR #301, opened 2015-08-14
Add API to clear table_cache

tzik:
PTAL.
Chrome holds multiple instances of LevelDB on its browser process. And they consume around 500kB for the read cache after 5 min browsing on a empty profile.
To drop them on a memory pressure, we'd like to add an API to leveldb::Cache.


tzik:
Taking down this PR to move this into the internal repository.

PR #303, opened 2015-09-18
Android NDK port for LevelDB.

K-Constantine:
for NDK-BUILD, an error saying " ‘fread_unlocked’ was not declared in this scope" always come up. a clean fix was to manually use c++ lock_guard on fread and fwrite functions.


cmumford:
Hi @K-Constantine:

Thanks for submitting, but at present leveldb is not adding ports to other platforms. More info at https://github.com/google/leveldb#contributing-to-the-leveldb-project

PR #304, opened 2015-09-25
Implement support for limited record lifetime

vadimskipin:
Proposed changes allows user to specify policy to drop expired entries on compaction.
Last seqno also stored in files meta data to allow expired files to be dropped without expensive compaction process.
Keeping seqno in files' meta data also allows us to completely skip old sstables on read.

### DIFF
--- a/db/db_impl.cc
+++ b/db/db_impl.cc
@@ -25,6 +25,7 @@
 #include "leveldb/status.h"
 #include "leveldb/table.h"
 #include "leveldb/table_builder.h"
+#include "leveldb/sequence_policy.h"
 #include "port/port.h"
 #include "table/block.h"
 #include "table/merger.h"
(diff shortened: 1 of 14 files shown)


vadimskipin:
I signed it!

vadimskipin:
I signed it!

bwzhang2011:
@cmumford, any update with such issue ? could it some way improve the compaction of leveldb ?

juliangruber:
what do you think about adding a slightly more low level and flexibly api which tells you when compaction is happening? Then you can go ahead and remove entries yourself, or do other things!

cmumford:
@googlebot rescan

cmumford:
@vadimskipin thx for the PR. Can you elaborate on why you would like this change? What problem does it solve?

cmumford:
This feature can be implemented in the application.

PR #306, opened 2015-09-28
[PosixEnv::FileLock] fix

AndreyG:
boost::interprocess::file_lock: lock -> try_lock, it corresponds to method Env::FileLock specification.

### DIFF
--- a/util/env_boost.cc
+++ b/util/env_boost.cc
@@ -402,8 +402,10 @@ class PosixEnv : public Env {
       boost::interprocess::file_lock fl(fname.c_str());
       BoostFileLock * my_lock = new BoostFileLock();
       my_lock->fl_ = std::move(fl);
-      my_lock->fl_.lock();
-      *lock = my_lock;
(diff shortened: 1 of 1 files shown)


AndreyG:
I signed it!

cmumford:
@AndreyG: This change is to env_boost.cc which is not part of this project. Please reopen (or resubmit) this request if I've overlooked something.

AndreyG:
OK, but what project is env_boost.cc part of?

cmumford:
Ahhh - I see. I totally missed about that old windows branch. I'm going to re-open this pull request, but I need to sync up with another dev to decide what to do. We're trying to only allow in tested code - and we're only testing the master branch on Linux right now.

cmumford:
@AndreyG:

I went ahead and merged this untested. The windows branch is old and not maintained. I'd like to add Windows support into master eventually, but we need a Windows continuous integration builder, and probably a better port for that to happen.

PR #307, opened 2015-09-30
Implement support for meta blocks

vadimskipin:
This is proposal for feature that allows users to store arbitrary meta blocks (stats or something) inside sstables:
- Low-level API includes TableBuilder::AddMetaBlock and Table::ReadMeta along with Table::NewMetaIterator.
- Mid-level API includes Version::VisitMeta that accepts visitor to visit all meta data blocks defined for this version. It also provides simple range-filter to visit only subset of sstables.
- High-level API not provided.

### DIFF
--- a/db/version_set.cc
+++ b/db/version_set.cc
@@ -12,6 +12,7 @@
 #include "db/memtable.h"
 #include "db/table_cache.h"
 #include "leveldb/env.h"
+#include "leveldb/meta.h"
 #include "leveldb/table_builder.h"
 #include "table/merger.h"
 #include "table/two_level_iterator.h"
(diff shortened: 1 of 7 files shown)


pablopitta:
|I signed it!|

Il 30/09/2015 21:52, googlebot ha scritto:

vadimskipin:
I signed it!

cmumford:
@googlebot rescan

cmumford:
Thanks for the PR @vadimskipin. We strive to keep the API as lean and simple as possible. I think the user can accomplish the same thing by partitioning the keyspace or possibly with a custom comparator.

PR #309, opened 2015-10-01
Implement support for Intel crc32 instruction (SSE 4.2)

vadimskipin:
Added support for hardware crc32c instruction.

### DIFF
--- a/build_detect_platform
+++ b/build_detect_platform
@@ -71,16 +71,25 @@ if [ "$CXX" = "g++" ]; then
     MEMCMP_FLAG="-fno-builtin-memcmp"
 fi
 
+SSE_FLAG=
+if [ "$CXX" = "g++" ]; then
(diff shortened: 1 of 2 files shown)


vadimskipin:
I signed it!

AGSaidi:
This patch breaks support for non-x86 systems as written

cmumford:
Yes, noticed the break. In the process of manually pulling this fix in support for non x86 (and older CPU's) has been fixed. The change is currently stalled while waiting on something unrelated, and will hopefully be unblocked soon.

cmumford:
Well it took forever (sorry), but we've finally landed this. It needed some changes, some of which could have been done by @vadimskipin, but a few tweaks were more easily made in the Google repo which is why this pull request isn't being merged the GitHub way.

Thanks for this pull request @vadimskipin.

PR #313, opened 2015-11-16
Automatically fallback to regular file IO if mmap fails.

jcoffland:
This change allows LevelDB to continue operating when virtual memory is exhausted.

Despite LevelDB's use of `MmapLimiter`, which hard codes the maximum number of open mmaps to 1000 for 64-bit systems, LevelDB can still run out of virtual memory for mmaps.  Rather than fail, this PR allows LevelDB to automatically fallback to regular file IO when  `mmap()` fails.  LevelDB already does this anyway when 1000 mmaps are reached.

With this change it I've also removed `MmapLimiter`, which is no longer necessary.  This has the following additional benefits:
- Improves performance by eliminating mutex locks in `MmapLimiter::Acquire()` and `MmapLimiter::Release()`.
- Makes mmaps possible on 32-bit systems. (`MmapLimiter` disallowed this.)
- Potentially allows 64-bit systems to use more mmaps when virtual memory is available.
- Simplifies the code and removes a hard coded limit.

I believe this is a low risk change.  The main logic changes are limited to the function `NewRandomAccessFile()` and removal of `class MmapLimiter` only affects `class PosixMmapReadableFile` and `NewRandomAccessFile()`.  Both of these changes are limited to the file `env_posix.cc`.  No function or class signatures which are accessed outside of `env_posix.cc` have been changed.

### DIFF
--- a/util/env_posix.cc
+++ b/util/env_posix.cc
@@ -312,13 +312,13 @@ class PosixEnv : public Env {
   virtual Status NewRandomAccessFile(const std::string& fname,
                                      RandomAccessFile** result) {
     *result = NULL;
-    Status s;
     int fd = open(fname.c_str(), O_RDONLY);
     if (fd < 0) {
-      s = IOError(fname, errno);
-    } else if (mmap_limit_.Acquire()) {
+      return IOError(fname, errno);
+    }
+    if (mmap_limit_.Acquire()) {
       uint64_t size;
-      s = GetFileSize(fname, &size);
+      Status s = GetFileSize(fname, &size);
       if (s.ok()) {
         void* base = mmap(NULL, size, PROT_READ, MAP_SHARED, fd, 0);
         if (base != MAP_FAILED) {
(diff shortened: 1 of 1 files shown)


jcoffland:
I signed it!

jcoffland:
Any chance this will get looked at?  This is a fix for an important bug.  leveldb crashes when you run out virtual memory otherwise.

jcoffland:
Anyone there?  Could really use some help with this.

cmumford:
@googlebot rescan

cmumford:
Thx for the PR @jcoffland. One advantage of using MmapLimiter (not just called Limiter) is that it allows the tests to exercise both the mmap and regular I/O paths. Can you preserve this?

jcoffland:
@cmumford  Nope.

jcoffland:
The handling of this PR is really disappointing.  After 3 years of being ignored, I get a request to do more work.  I declined because there's no guarantee my work won't be ignored again.  This PR fixes an important bug in LevelDB and should have been merged a long time ago.

pwnall:
@jcoffland I'm sorry you feel this way.

Regarding timing -- We're focusing on what we expect to make the biggest improvements for the use cases in which we're deploying LevelDB.

Regarding testing -- We're running LevelDB on billions of devices, via Chrome. Correctness is really important to us, and adequate test coverage is not negotiable.

yujiapingyu:
Why I can't get the Value of a exsit Key when I set  maximum number of open mmaps to 0 ？

PR #316, opened 2015-11-19
Port leveldb to windows.

aloukissas:


andychase:
Bitcoin has had leveldb windows support for awhile. I'm not sure how this implementation is different, but note that this patch: https://github.com/bitcoin/leveldb/pull/9 came through recently that fixed a lot of corruption issues with the bitcoin implementation.

Hope this helps this project.

aloukissas:
@andychase thanks for the pointer. I'll look into this.

ralphtheninja:
> Bitcoin has had leveldb windows support for awhile. I'm not sure how this implementation is different, but note that this patch: bitcoin/leveldb#9 came through recently that fixed a lot of corruption issues with the bitcoin implementation.
> 
> Hope this helps this project.

Oh nice one. It would be great to get that merged into the upstream repo!

cmumford:
Thanks for the pull request. Unfortunately supporting Windows is not on our current agenda. Please see https://github.com/google/leveldb#contributing-to-the-leveldb-project for a slightly better explanation as to why.

Regarding the corruption issues. If the native Win32 I/O fixed any corruption issues then that would indicate a missing flush or sync in leveldb, and the correct fix should be in leveldb proper and not in the Env. Also, Chrome (which uses leveldb) switched from streams (i.e. fopen, fwrite) to native Windows I/O calls, and saw no change in corruption, so I suspect that any bug (or fix) lives elsewhere.

andychase:
Thanks @cmumford. From cursory overview it appears they used win32 memory mapping, which is not robust against computer crashes; so no upstream patching need.

Thanks for the input.

topilski:
Good day may be this implementation of fsync will be interesting for you:
```
static ssize_t fsync(int fd) {
    HANDLE h = (HANDLE) _get_osfhandle (fd);
    DWORD err;

    if (h == INVALID_HANDLE_VALUE) {
      errno = EBADF;
      return -1;
    }

    if (!FlushFileBuffers (h)) {
      /* Translate some Windows errors into rough approximations of Unix
      * errors.  MSDN is useless as usual - in this case it doesn't
      * document the full range of errors.
      */
      err = GetLastError ();
      switch (err) {
      case ERROR_ACCESS_DENIED:
        /* For a read-only handle, fsync should succeed, even though we have
        no way to sync the access-time changes.  */
        return 0;

      /* eg. Trying to fsync a tty. */
      case ERROR_INVALID_HANDLE:
        errno = EINVAL;
        break;

      default:
        errno = EIO;
      }
      return -1;
    }

    return 0;
}

static ssize_t pread (int fd, void *buf, size_t count, off_t offset) {
  if (lseek(fd, offset, SEEK_SET) != offset)
    return -1;
  return read(fd, buf, count);
}
```

we build leveldb on Windows via msys2 and mingw and it is worked.

Issue #320, opened 2015-11-27
Compaction causes data inconsistency when using snapshots


specialforest:
When using snapshots, background compaction causes data inconsistency. Existing records can became missing or previously existed records or their values can appear again. Test program that reproduces the issue https://gist.github.com/specialforest/fa136b660bdd672e3d4b

richcole:
The error occurs when leveldb does a compaction while there are several records with the same key but different sequence numbers. This can occur when snapshots are in play. The compaction algorithm can choose a file such that it will push the latest record from level i to level i+1, then a following get will retrieve the older record from level i which is incorrect. I see two ways to fix: (i) always compact a file set that is closed with respect records with the same key, e.g. if files f1 and f2 have records with key k1 then compact f1 and f2 together, or (ii) when considering f1 for compaction check that no older record of any k in f1 exists at level i, it is sufficient to check the boundary elements of the file f1 I believe, and if the criteria isn't met to try to compact the previous file (I assume you sort based on key and sequence number with increasing sequence numbers being later in the sequence of files at a level).

richcole-at-amazon:
In case it isn't obvious from the pretty graphics above included by github there is a patch for this bug here:

https://github.com/google/leveldb/pull/339

PR #339, opened 2016-01-29
Fix snapshot compaction bug

richcole-at-amazon:
Closes google/leveldb#320

During compaction it was possible that records from a block b1=(l1,u1)
would be pushed down from level i to level i+1. If there is a block
b2=(l2,u2) at level i with k1 = user_key(u1) = user_key(l2) then
a subsequent search for k1 will yield the record l2 which has a smaller
sequence number than u1 because the sort order for records sorts
increasing by user key but decreaing by sequence number.

This change add a call to a new function AddBoundaryInputs to
SetupOtherInputs. AddBoundaryInputs searches for a block b2 matching the
criteria above and adds it to the set of files to be compacted. Whenever
AddBoundaryInputs is called it is important that the compaction fileset
in level i+1 (known as c->inputs_[1] in the code) be recomputed. Each
call to AddBoundaryInputs is followed by a call to GetOverlappingInputs.

SetupOtherInputs is called on both manual and automated compaction
passes. It is called for both level zero and for levels greater than 0.

Testing Done:

  A test program (issue320_test) was constructed that performs mutations
  while snapshots are active. issue320_test fails without this bug fix
  after 64k writes. It passes with this bug fix. It was run with 200M
  writes and passed.

  Unit tests were written for the new function that was added to the
  code. Make check was run and seen to pass.

Signed-off-by: Richard Cole richcole@amazon.com

### DIFF
--- a/db/version_set_test.cc
+++ b/db/version_set_test.cc
@@ -172,7 +172,153 @@ TEST(FindFileTest, OverlappingFiles) {
   ASSERT_TRUE(Overlaps("600", "700"));
 }
 
-}  // namespace leveldb
+void AddBoundaryInputs(const InternalKeyComparator &icmp,
+                       const std::vector<FileMetaData *> &levelFiles,
+                       std::vector<FileMetaData *> *compactionFiles);
+
+class AddBoundaryInputsTest {
+ public:
+  std::vector<FileMetaData *> levelFiles_;
+  std::vector<FileMetaData *> compactionFiles_;
+  std::vector<FileMetaData *> allFiles_;
+  InternalKeyComparator icmp_;
+
+  AddBoundaryInputsTest() : icmp_(BytewiseComparator()){};
+
+  ~AddBoundaryInputsTest() {
+    for (size_t i = 0; i < allFiles_.size(); ++i) {
+      delete allFiles_[i];
+    }
+    allFiles_.clear();
+  };
+
+  FileMetaData *CreateFileMetaData(uint64_t number, InternalKey smallest,
+                                   InternalKey largest) {
+    FileMetaData *f = new FileMetaData();
+    f->number = number;
+    f->smallest = smallest;
+    f->largest = largest;
+    allFiles_.push_back(f);
+    return f;
+  }
+};
+
+TEST(AddBoundaryInputsTest, TestEmptyFileSets) {
+  AddBoundaryInputs(icmp_, levelFiles_, &compactionFiles_);
+  ASSERT_TRUE(compactionFiles_.empty());
+  ASSERT_TRUE(levelFiles_.empty());
+}
+
+TEST(AddBoundaryInputsTest, TestEmptyLevelFiles) {
+  FileMetaData *f1 =
+      CreateFileMetaData(1, InternalKey("100", 2, kTypeValue),
+                         InternalKey(InternalKey("100", 1, kTypeValue)));
+  compactionFiles_.push_back(f1);
+
+  AddBoundaryInputs(icmp_, levelFiles_, &compactionFiles_);
+  ASSERT_EQ(1, compactionFiles_.size());
+  ASSERT_EQ(f1, compactionFiles_[0]);
+  ASSERT_TRUE(levelFiles_.empty());
+}
+
+TEST(AddBoundaryInputsTest, TestEmptyCompactionFiles) {
+  FileMetaData *f1 =
+      CreateFileMetaData(1, InternalKey("100", 2, kTypeValue),
+                         InternalKey(InternalKey("100", 1, kTypeValue)));
+  levelFiles_.push_back(f1);
+
+  AddBoundaryInputs(icmp_, levelFiles_, &compactionFiles_);
+  ASSERT_TRUE(compactionFiles_.empty());
+  ASSERT_EQ(1, levelFiles_.size());
+  ASSERT_EQ(f1, levelFiles_[0]);
+}
+
+TEST(AddBoundaryInputsTest, TestNoBoundaryFiles) {
+  FileMetaData *f1 =
+      CreateFileMetaData(1, InternalKey("100", 2, kTypeValue),
+                         InternalKey(InternalKey("100", 1, kTypeValue)));
+  FileMetaData *f2 =
+      CreateFileMetaData(1, InternalKey("200", 2, kTypeValue),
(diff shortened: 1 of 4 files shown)


vonnyfly:
Why does this patch not to be merged to the master?
This is a very serious defect.
@richcole-at-amazon @jeffreyadean 

Is there any unknown background behind this?

vonnyfly:
@ghemawat 
@costan

kylezh:
@cmumford @pwnall @ghemawat 

Please take some time to a look at this patch. LevelDB is a widely used library and this is really a very critical bugfix.

I can reproduce this bug on master branch(9ce3051).

And I think this is the same problem of #375 

Thanks!

cmumford:
Just sent an email to @richcole-at-amazon to see if he is able to sign the [Google CLA](https://cla.developers.google.com/.).

Makefile
```diff
@@ -175,6 +176,9 @@ issue178_test: issues/issue178_test.o $(LIBOBJECTS) $(TESTHARNESS)
 issue200_test: issues/issue200_test.o $(LIBOBJECTS) $(TESTHARNESS)
 	$(CXX) $(LDFLAGS) issues/issue200_test.o $(LIBOBJECTS) $(TESTHARNESS) -o $@ $(LIBS)
 
+issue320_test: issues/issue320_test.o $(LIBOBJECTS) $(TESTHARNESS)
```
cmumford: @richcole-at-amazon the Makefile has been removed since you first submitted this and replaced with CMakeLists.txt (for cmake). Can you make a one-line addition to that file for this new test?

db/version_set_test.cc
```diff
@@ -172,7 +172,134 @@ TEST(FindFileTest, OverlappingFiles) {
...
+  AddBoundaryInputs(icmp_, levelFiles_, &compactionFiles_);
+  ASSERT_EQ(3, compactionFiles_.size());
+  ASSERT_EQ(f1, compactionFiles_[0]);
+  ASSERT_EQ(f3, compactionFiles_[1]);
+  ASSERT_EQ(f2, compactionFiles_[2]);
+}
+
+TEST(AddBoundaryInputsTest, TestDisjoinFilePointers) {
+  FileMetaData *f1 = CreateFileMetaData(1, InternalKey("100", 6, kTypeValue), InternalKey(InternalKey("100", 5, kTypeValue)));
+  FileMetaData *f2 = CreateFileMetaData(1, InternalKey("100", 2, kTypeValue), InternalKey(InternalKey("300", 1, kTypeValue)));
```
cmumford: f2 and f3 are unused in this test. Can you delete these local variables or use if this is an oversight?

issues/issue320_test.cc
```diff
@@ -0,0 +1,130 @@
...
+    ++i;
+  }
+  while (i < sizeof(bytes)) {
+    bytes[i] = 'a' + random(26);
+    ++i;
+  }
+  return string(bytes, sizeof(bytes));
+}
+
+int main(int argc, char** argv) {
```
cmumford: This test is fine, but there were two problems I noticed when running on the Google internal infrastructure. 1. Our tests detected leaking snapshots, and 2) We cannot create local files - they need to be put in a temporary directory. I made those two changes as well as refactoring this test to use testutil.h - which is similar to googletest.

I uploaded my refactoring to https://github.com/cmumford/leveldb/commit/4b72cb14f8da2aab12451c24b8e205aff686e9dc. You're welcome to copy my version of the file over if you want.

richcole-at-amazon:
Thanks Chris. I'll look at making these changes and updating the pull request. I'm going to try to carve out some time on Thursday.

db/version_set_test.cc
richcole-at-amazon: removed f2 and f3, then renamed variables to make the numbers sequential.

issues/issue320_test.cc
```diff
@@ -0,0 +1,139 @@
...
+  }
+  return string(bytes, sizeof(bytes));
+}
+
+}  // namespace
+
+class Issue320 { };
+
+TEST(Issue320, Test) {
+  srandom(0);
```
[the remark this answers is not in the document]
richcole-at-amazon: Sure should be straight forward. Will update.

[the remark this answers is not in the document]
cmumford: Oh, one more optional thing. I failed to point out, in your PR, code that didn't fully conform to the [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html). I won't require it for this PR and will fix up the formatting after merging this change. However, if you want Git blame to more accurately identify you as the author of certain lines you're welcome to pull over my formatting fixes from https://github.com/cmumford/leveldb/commit/783fcffd74a4e4c5f5398bb2ea463a6def6e747c. If not then I'll just land them in a post merge cleanup.

ruish:
hi Chris @cmumford,

I was trying to patch your change https://github.com/cmumford/leveldb/commit/48a9a9a0f9f39d9df02e0c1e14f647f8a7e96b23 to my local repo.

I first build and test the issue320_test.cc without patching other part of the fix. To my surprise, the test was passed. I expected it to fail without the fix.

Could you please verify that? Should the issue320_test pass without other part of the fix?

Thanks,

Rui

ruish:
BTW, my local repo is one based on this commit: https://github.com/google/leveldb/commit/7b945f200339aa47c24788d3ee9910c09c513843

cmumford:
Thanks @ruish - I introduced an error in the test when refactoring. I've uploaded a fix to https://github.com/cmumford/leveldb/commit/9c8d57c881be5d42a72a8e57f9c165b602716bd5.

cmumford:
Note. I'm trying to configure Copybara to accept this PR. Please ignore any of those errors.

ruish:
Hi Chris @cmumford ,

LGTM!

I patched your fix in my local repo and verified that the test is passed/failed with/without the original fix, which meets the expectation.

Thanks!

Rui

jsolman:
Using LevelDB built from the master branch of this repository, I was hitting this issue fairly consistently when syncing the Neo (https://neo.org) MainNet Blockchain from the latest chain file available at https://sync.ngd.network. In certain instances where compaction would occur, keys were becoming missing. I have not encountered the issue since using a build that incorporates the changes from this PR.

PR #345, opened 2016-02-10
Remove const from Delete

eyolfson:
The Delete method clearly modifies the input list and is confusing being marked as const. In addition to modification through pointers, s is deleted. Perhaps New and Delete should be refactored into proper constructors/destructors.

### DIFF
--- a/db/snapshot.h
+++ b/db/snapshot.h
@@ -50,7 +50,7 @@ class SnapshotList {
     return s;
   }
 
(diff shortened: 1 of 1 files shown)


cmumford:
@googlebot rescan

cmumford:
Hi @eyolfson. can you fix the failed cast identified in https://travis-ci.org/google/leveldb/jobs/108317197?

cmumford:
Closing due to inactivity.

PR #346, opened 2016-02-10
Remove const from Options or refactor

eyolfson:
See changes with comment. The options variable's env field is transitively modified through a const reference, which is visible after the method completes.


cmumford:
@googlebot rescan

cmumford:
Thanks for the pull request. In this case Options does not own the Env, but merely has a pointer to it. In most uses options.env refers to the a Env singleton returned by Env::Default(). If Options owned the env then your suggested change would be correct.

PR #350, opened 2016-02-24
Update db_impl.cc

liaochengcai:
Read/Write requests are blocked shortly when leveldb is compacting. we found that the unlink 
function costs too much time in DeleteObsoleteFiles().  (unlink(sstableFile) cost 50ms , sstable 
size=256m), while unlink function doesn't need to acquire a lock.What's more, in 
DeleteObsoleteFiles(), unlink is called in a for loop,  which causes a much worse -- during this
 time, requests are blocked util the lock is released. It will improve levaldb's availability by adding 
unlock/lock before/after env_->DeleteFile() call. —— R&D Team of YY.inc Entertainment dept.

### DIFF
--- a/db/db_impl.cc
+++ b/db/db_impl.cc
@@ -264,7 +264,9 @@ void DBImpl::DeleteObsoleteFiles() {
         Log(options_.info_log, "Delete type=%d #%lld\n",
             int(type),
             static_cast<unsigned long long>(number));
+        mutex_.Unlock();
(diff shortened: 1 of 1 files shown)


liaochengcai:
I signed it!

cmumford:
@googlebot rescan

cmumford:
Thanks for the PR @liaochengcai. A different solution to this problem just landed in 4cb80b7ddce6ff6089b15d8cfebf746fc1572477.

PR #352, opened 2016-03-02
Call cleanups from last registered to first

codeeply:
It seems more reasonable that cleanups are called from the last registered one to the first one,
when destruct.

### DIFF
--- a/table/iterator.cc
+++ b/table/iterator.cc
@@ -13,13 +13,13 @@ Iterator::Iterator() {
 
 Iterator::~Iterator() {
   if (cleanup_.function != NULL) {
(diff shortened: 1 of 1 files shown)


codeeply:
I signed it!

cmumford:
@googlebot rescan

cmumford:
Thanks for the PR @codeeply. I think that this might have been a better way to implementation this from the start. However, reversing the order of callback execution could negatively impact existing leveldb users, so it's safer to keep with the existing solution. If you need a reverse cleanup mechanism for Iterator then you can implement your own in application code.

Sorry for the long delay in considering your PR. Please reopen if you want to continue this discussion.

Issue #361, opened 2016-04-02
uint64_t is not a member of "std"


mrmiywj:
Caused by  [LC107](https://github.com/google/leveldb/blob/master/db/recovery_test.cc#L107) and 
[LC110](https://github.com/google/leveldb/blob/master/db/recovery_test.cc#L110)

mrmiywj:
Arch linux64  GCC 5.3.0

cmumford:
Fixed in ea992b4

PR #362, opened 2016-04-02
fix uint64

mrmiywj:
fix [#361](https://github.com/google/leveldb/issues/361)


mrmiywj:
I signed it!

cmumford:
Addressed in pull request nw362#354

PR #367, opened 2016-04-13
Fix LevelDB build when asserts are enabled in release builds.

jabdelmalek:
BUG=https://bugs.chromium.org/p/chromium/issues/detail?id=603166

### DIFF
--- a/table/format.cc
+++ b/table/format.cc
@@ -30,9 +30,7 @@ Status BlockHandle::DecodeFrom(Slice* input) {
 }
 
 void Footer::EncodeTo(std::string* dst) const {
-#ifndef NDEBUG
   const size_t original_size = dst->size();
-#endif
   metaindex_handle_.EncodeTo(dst);
   index_handle_.EncodeTo(dst);
   dst->resize(2 * BlockHandle::kMaxEncodedLength);  // Padding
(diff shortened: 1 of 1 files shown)


cmumford:
@jabdelmalek How are you getting assert() to be defined if NDEBUG is not defined? I'm assuming that assert() is trying to use original_size which was never defined, but can't see how that happens.

jabdelmalek:
@cmumford This happens in Chrome's release-with-asserts builds that are used on the commit queue.  Those configs define DCHECK_ALWAYS_ON.

Also logging.h in chromium redefines assert to be DLOG_ASSERT:
https://code.google.com/p/chromium/codesearch#chromium/src/base/logging.h&l=741

Which for the release-with-asserts build ends up needing original_size.

cmumford:
@jabdelmalek: OK - so Chromium is defining assert() even though NDEBUG is defined - that seems wrong. I know this is a minor thing, but I'm trying to find a way to allow `-Werror -Wunused-variable` to be used without breaking the build - and right now you can do that and get a clean build. With this change the release build will fail (with those flags).

The sole purpose of NDEBUG is to disable the assert macro, and Chrome is ignoring that.

jabdelmalek:
@cmumford I had tried the release build and it passed. I actually tested all 3 build configs in the standalone LevelDB build and also all flavors of Chromium (debug, release, release-with-asserts).

The linked chromium bug has more information about the bug. Basically without the fix, we are having changes land through the commit queue that break the main waterfall. Note that we have been running release-with-asserts builds from the start of the commit queue, as a compromise between getting coverage from asserts while not having super slow tests.

jabdelmalek:
@cmumford per our offline conversation, I can now repro your failure after adding the -Werror -Wunused-variable flags. I've updated the change to fix this. Thanks.

cmumford:
leveldb does have a precedent for this - look in `AcquireLoad()`. Can you add a comment above `(void)original_size;`? After that I think it will be OK.

jabdelmalek:
Done, thanks.

PR #369, opened 2016-04-24
Add missing status assignment

kezhuw:
Old code use `s.ok()` to guard `TableBuilder::Finish`, but `s` was ok
before building and was not reassigned from `TableBuilder::status` after
building. `s` evaluates to true always in the `if` guard.


cmumford:
@googlebot rescan

kezhuw:
@cmumford  Next line is a `s.ok` check already.

cmumford:
@kezhuw The file changed since your initial pull request in ea0a7586b8615fd39c6b8f5a8a21a1f242129c2f and that check was removed.

kezhuw:
@cmumford Thanks for pointing out, I didn't notice that. I have revert ea0a758 to keep origin logic. Personal, I think both fixes are correct in functionality. But adding a status assignment seems more likely be the original intention.

PR #370, opened 2016-04-29
provide a easy way to check with error

makunterry:
Keep error msg stored in static, that caller won't need to free buffer for it anymore.


makunterry:
errptr is address of a pointer, we just change the pointer value. :)
@doitliao

cmumford:
@googlebot rescan

cmumford:
Thank you for the PR @makunterry. This change would require a change to the public C API which would break existing clients. Also this implementation isn't threadsafe.

PR #373, opened 2016-05-09
Reset grandparents overlapped bytes after current output has been fin…

kezhuw:
…ished due to file size limitation

If current ouput is finished due to file size limitation, possibilities
exist that we hit grandparents overlapped bytes limitation after writing
fewer entries. So we need to restart grandparents overlapped bytes if we
start building a new output for reasons other than overlapped bytes with
grandparents.

This commit adds code to restart grandparents overlapped counting after
current output has been successfully finished due to file size
limitation.

I think the best place to restart grandparents overlapped counting
should be DBImpl::OpenCompactionOutputFile, thus we don't need to
remember to do this every time after we stop or start a output.


cmumford:
@googlebot rescan

PR #378, opened 2016-05-27
Reuse previously declared local variable to prevent -Wshadow warnings

paveljanik:


paveljanik:
Technically, I do not contribute to your project. I have removed something from the project. Do you really want signed CLA even for removals? ;-)

I have removed 11 characters...

paveljanik:
No need to rebase, the problem is fixed by other changes in the code!

PR #392, opened 2016-07-10
Some cleanup of the util/coding library

jonnyhsy:
1) Removing the obselete api: GetLengthPrefixedSlice(const char*, const char*, Slice*)
as it's not referenced anywhere.
2) A small implementation tweak of GetVarint32/GetVarint64 api:advancing the
input slice past the parsed value using Slice::remove_prefix() instead of
reinitializing *input_slice object.
3) coding_test.cc update.

Test: make check; all tests passed.

### DIFF
--- a/util/coding.cc
+++ b/util/coding.cc
@@ -135,7 +135,7 @@ bool GetVarint32(Slice* input, uint32_t* value) {
   if (q == NULL) {
     return false;
   } else {
-    *input = Slice(q, limit - q);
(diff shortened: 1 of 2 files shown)


jonnyhsy:
I signed it!

bwzhang2011:
@googlebot, would you mind leaving some message that when next version will be released since 1.18 ?

cmumford:
@googlebot rescan

PR #393, opened 2016-07-15
Add compiler flag to fix build in cygwin

antlad:


antlad:
I signed it!

cmumford:
Thx for the PR @antlad. We switched to CMake recently which hopefully works for cygwin.

PR #397, opened 2016-08-04
Fix error in code commenting

xiyanxiyan10:
I found it when I view the code of leveldb。


juliangruber:
in case you thought this is a typo, "iff" stands for "if and only if", see https://en.wikipedia.org/wiki/If_and_only_if

xiyanxiyan10:
I get it,Thanks.

PR #399, opened 2016-08-05
bug fix for string key case

codeeply:
It will coredump, with error 'basic_string::_S_construct null not valid'.

```c++
#include <iostream>
#include "skiplist.h"
#include "util/arena.h"

using namespace std;
using namespace leveldb;

struct StringComparator {
  int operator()(const string& a, const string& b) const {
      return a < b;
  }
};

int main()
{
    StringComparator comp;
    Arena arena;

    SkipList<string, StringComparator> tmp(comp, &arena);
    tmp.Insert(string("a"));

    cout << tmp.Contains(string("b")) << endl;
    cout << tmp.Contains(string("a")) << endl;

    cout << "test skiplist" << endl;

    return 0;
}
```

### DIFF
--- a/db/skiplist.h
+++ b/db/skiplist.h
@@ -325,7 +325,7 @@ template<typename Key, class Comparator>
 SkipList<Key,Comparator>::SkipList(Comparator cmp, Arena* arena)
     : compare_(cmp),
       arena_(arena),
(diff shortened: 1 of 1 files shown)


codeeply:
I signed it!

cmumford:
@googlebot rescan

PR #409, opened 2016-09-19
documentation: fix incorrect log file default size

mindprince:
According to the rest of the documentation, a new sstable is created from the memtable when the log file grows to more than 1MB in size. 4MB is the threshold which triggers a level-0 compaction.


mindprince:
I signed it!

cmumford:
@googlebot rescan

cmumford:
I believe 4MB is correct as this text refers to Options::write_buffer_size. Also, impl.html was converted to Markdown and is now impl.md.

Issue #412, opened 2016-10-16
libleveldb shared library name does not match platform conventions on macOS


ryandesign:
Hello, I'm a developer with the MacPorts package management system.

The name of the libleveldb shared library, and its embedded install_name, does not match the conventions of macOS.

For example, installing leveldb 1.18 via MacPorts, these files are installed:

```
$ ls -l /opt/local/lib/libleveldb*
-rw-r--r--  1 root  wheel  991072 Oct 16 16:01 /opt/local/lib/libleveldb.a
lrwxr-xr-x  1 root  wheel      21 Oct 16 16:01 /opt/local/lib/libleveldb.dylib -> libleveldb.dylib.1.18
lrwxr-xr-x  1 root  wheel      21 Oct 16 16:01 /opt/local/lib/libleveldb.dylib.1 -> libleveldb.dylib.1.18
-rwxr-xr-x  1 root  wheel  625960 Oct 16 16:01 /opt/local/lib/libleveldb.dylib.1.18
```

And the shared libraries have this install_name:

```
$ otool -L /opt/local/lib/libleveldb.dylib
/opt/local/lib/libleveldb.dylib:
    /opt/local/lib/libleveldb.dylib.1 (compatibility version 0.0.0, current version 0.0.0)
    /opt/local/lib/libsnappy.1.dylib (compatibility version 5.0.0, current version 5.0.0)
    /opt/local/lib/libtcmalloc.4.dylib (compatibility version 7.0.0, current version 7.6.0)
    /usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 120.1.0)
    /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1226.10.1)
```

On macOS, the library version number does not go after the filename extension; it goes before. So the library should not be "libleveldb.dylib.1"; it should be "libleveldb.1.dylib".

This problem was noticed by the Homebrew package management system as well in https://github.com/Homebrew/legacy-homebrew/issues/14440. There was a comment there that an upstream issue should be submitted, but I don't know if one ever was, so I'm submitting this one now.

I have patched the build system to fix this and will submit a pull request shortly.

cmumford:
The build process has switched from make to CMake and no longer creates versioned library symbolic links.

PR #413, opened 2016-10-16
Fix libleveldb shared library install_name on macOS

ryandesign:
This resolves issue #412, so that libleveldb gets installed on macOS as "libleveldb.1.dylib" instead of "libleveldb.dylib.1".

A new variable `PLATFORM_SHARED_BASE` is introduced which contains "libleveldb" on macOS and "libleveldb.so" on other systems.

The meaning of the `PLATFORM_SHARED_EXT` variable is changed to mean the extension that goes after the version number, including the period. This is ".dylib" on macOS and empty on other systems.

`PLATFORM_SHARED_EXT` previously had a second meaning: if empty, it meant no shared library should be built. This second meaning is now moved to a new Boolean variable, `PLATFORM_SHARED`.

I have only tested on OS X 10.11 El Capitan but it should work on any version of Mac operating system. It is my intention that this should not change the behavior on non-Mac operating systems but I don't have any available to test with, so please test on other systems before merging.

I have submitted a CLA as "Ryan Schmidt".


ryandesign:
I've rebased this PR. Has anyone tested it on non-Mac systems? Do I need to do anything else before this can be merged?

pwnall:
Thank you very much for the PR, and I'm sorry we haven't had the bandwidth to react sooner!

While I could look into merging this, it'd have a very short life, as we're going to switch to CMake for the next release. Work in progress at https://github.com/pwnall/leveldb/tree/cmake

If you happen to have some time, I'd appreciate it if you'd try out that branch and comment here on whether it works or what's missing.

ryandesign:
Thanks! That's good news. As a maintainer of the leveldb package for MacPorts, I do prefer packages that use popular build systems, since their behaviors and quirks are well-known.

I tried your branch, and `make` succeeded, but `make install` eventually failed with:

```
CMake Error at cmake_install.cmake:82 (file):
  file INSTALL cannot find
  "/opt/local/var/macports/build/_Users_rschmidt_macports_macports-ports-svn-trunk_databases_leveldb/leveldb/work/leveldb-bc700bd394d67be2a633707d2ad9b898dae8d79c/LeveldbConfig.cmake".
```

It also looks like it only built static libraries (libleveldb.a, libleveldb_bench_base.a, libleveldb_memenv.a, libleveldb_test_base.a). The old build system built both static libraries and dynamic (shared) libraries, and this bug report pertains to the dynamic libraries.

pwnall:
Thank you for the feedback! I've updated the branch.

As far as I know, this is how CMake works -- you can build shared libraries in a separate pass by adding -DBUILD_SHARED_LIBS=ON to the CMake command line.

The CMake config here follows the same structure as the one for google/snappy, as I pieced both of them together. If MacPorts packages snappy, it might help to see how that package handles the static/shared libraries issue.

ryandesign:
> As far as I know, this is how CMake works -- you can build shared libraries in a separate pass by adding -DBUILD_SHARED_LIBS=ON to the CMake command line.

Having to run a second pass of cmake (and make and make install?) is really crappy! We'll probably settle for only having dynamic libraries then. I didn't realize `-DBUILD_SHARED_LIBS=ON` was a standard flag that would work with any cmake project. I also apparently don't know how to display all of the cmake flags that a project recognizes (like `./configure --help` for autotools).

> The CMake config here follows the same structure as the one for google/snappy, as I pieced both of them together. If MacPorts packages snappy, it might help to see how that package handles the static/shared libraries issue.

Ever since snappy switched to cmake, it has been quite broken in MacPorts. It too stopped installing a dynamic library when we updated, and if dynamic libraries are built, they have the wrong version information; see https://trac.macports.org/ticket/55583.

ryandesign:
> Thank you for the feedback! I've updated the branch.

Thanks, I'm able to build it now, from e98e51202e7810a60b6ca70c2465ae86b3e5fce9.

Could you add library versioning information? Currently you have:

```
$ otool -L /opt/local/lib/libleveldb.dylib 
/opt/local/lib/libleveldb.dylib:
	/opt/local/lib/libleveldb.dylib (compatibility version 0.0.0, current version 0.0.0)
	/opt/local/lib/libtcmalloc.4.dylib (compatibility version 10.0.0, current version 10.1.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 307.5.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1238.60.2)
```

Since this is a library that other projects consume, the library should advertise major and minor version numbers to convey when your updates are or are not compatible with previous releases. The Makefile said:

```
# Update db.h if you change these.
SHARED_VERSION_MAJOR = 1
SHARED_VERSION_MINOR = 20
```

So I would expect to see output more like this instead:

```
$ otool -L /opt/local/lib/libleveldb.1.dylib 
/opt/local/lib/libleveldb.1.dylib:
	/opt/local/lib/libleveldb.1.dylib (compatibility version 20.0.0, current version 20.0.0)
	/opt/local/lib/libtcmalloc.4.dylib (compatibility version 10.0.0, current version 10.1.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 307.5.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1238.60.2)
```

I don't know if that compatibility version is correct. If library major version 1 minor version 20 is compatible with earlier versions of library major version 1, then change the compatibility version to the earliest minor version with which it is compatible.

include/leveldb/db.h says:

```
// Update Makefile if you change these
static const int kMajorVersion = 1;
static const int kMinorVersion = 20;
```

This should be changed to mention editing CMakeLists.txt instead of Makefile.

ryandesign:
The documentation says that on macOS, cmake uses `VERSION` as the current version (i.e. minor version) and `SOVERSION` as the compatibility version. It apparently does not provide a way to specify the major version?

pwnall:
@ryandesign I've been busy pushing CMake and some internal cleanups through our internal repository. I don't have a good answer on versioning and CMake. If you happen to figure that out before I do, please comment here.

cmumford:
This was fixed in https://github.com/google/leveldb/commit/37300aa54b8256dd2edfd504942eb2bd20823647 and tweaked in https://github.com/google/leveldb/commit/09fa8868dbe0cb2701f0560c59ebb63cc17f1271.

PR #417, opened 2016-10-22
Remove extra new lines from logging

mnaser:
The built-in logger strips new lines if they are included, however,
in the case when another application is consuming the logs (such as
Ceph), it will get some log lines with new lines and some without.

This patch removes a large majority of new lines in an attempt to
make logging always consistently deliver messages without new lines.

### DIFF
--- a/db/db_impl.cc
+++ b/db/db_impl.cc
@@ -262,7 +262,7 @@ void DBImpl::DeleteObsoleteFiles() {
         if (type == kTableFile) {
           table_cache_->Evict(number);
         }
-        Log(options_.info_log, "Delete type=%d #%lld\n",
+        Log(options_.info_log, "Delete type=%d #%lld",
             int(type),
             static_cast<unsigned long long>(number));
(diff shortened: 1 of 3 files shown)


mnaser:
HI everyone, I'd like to give this a small bump.

cmumford:
@googlebot rescan

cmumford:
@mnaser thx for the PR and sorry for the delay. The current logger (`posix_logger.h`) doesn't strip \n's, but it will add them if the string doesn't end with one. I didn't look at an older version of that file at the time of your PR. I think the current version should be compatible with Ceph. WDYT?

mnaser:
https://github.com/mnaser/leveldb/blob/master/util/posix_logger.h#L80-L83

It seemed to be the case back then as well.. but I guess somehow the logs still end up with multiple lines.

cmumford:
What platform are you running on?

mnaser:
While it seemed to be a while ago, this was a Ceph deployment using ceph-ansible. I haven't dug in yet to see if this is still happening however, sorry.

PR #422, opened 2016-11-07
Fix macOS 10.12 build warning

droark:
OSMemoryBarrier() has been deprecated as of macOS 10.12. Compile it only if on a version where it’s not deprecated.


droark:
Just FYI, this is untested beyond confirming that it compiles without warnings under 10.12. I'd imagine it'll be fine elsewhere. I just haven't tested. If I'm wrong, I'm happy to mod the patch as needed.

Thanks.

spencerlievens:
ACK

droark:
Update: I finally ran all the tests and `db_bench` in static mode, along with `db_bench` in shared mode. They all passed (tests) or had what appeared to be reasonable numbers (`db_bench`).

Thanks.

tmm1:
Isn't it necessary to replace this with something else that implements the same barrier?

From the deprecation message, it seems `std::atomic_thread_fence()` is recommended:

```
./port/atomic_pointer.h:59:3: warning: 'OSMemoryBarrier' is deprecated: first deprecated in macOS 10.12 - Use std::atomic_thread_fence() from <atomic>
      instead [-Wdeprecated-declarations]
  OSMemoryBarrier();
  ^
/usr/include/libkern/OSAtomicDeprecated.h:749:9: note: 'OSMemoryBarrier' has been explicitly marked deprecated here
void    OSMemoryBarrier( void );
        ^
1 warning generated.
```

droark:
@tmm1 - Fair point. I have noticed that a couple of similar warnings were fixed in a different manner that followed what the compiler prescribed. I'll look into a solution that maintains the barrier.

Thanks.

droark:
Added [atomic_thread_fence()](http://en.cppreference.com/w/cpp/atomic/atomic_thread_fence), as suggested by the compiler. I chose [sequentially-consistent ordering](http://en.cppreference.com/w/cpp/atomic/memory_order#Sequentially-consistent_ordering), as that seemed like the most logical choice. I'm happy to change this as the LevelDB team sees fit, though.

Re-ran all the tests as mentioned above. Everything seemed fine.

theuni:
See #449, which is IMO the preferable fix.

droark:
@theuni - Thanks. I'll leave this open just in case but, after thinking about it, I think I prefer the fix @sipa put together.

sipa:
Is there any chance that someone would compile on/for OS X 10.12, but with a c++ compiler that doesn't have c++11 atomics? If so, we should probably get both patches in, so that atomics are preferred, but a deprecation warning for MemoryBarrier is still shown.

droark:
Fixed [here](https://github.com/google/leveldb/commit/594cc987af2e0af6417c4ac2b947ee8cdad59e5e). Closing this out.

PR #448, opened 2017-02-10
to use stdint.h macros which declares SIZE_MAX, comment to entry format is error

andyli029:
‘SIZE_MAX’ was not declared in when compiling with OPT ?= -g2 in Makefile,
Compiling on Linux with g++ -D__STDC_LIMIT_MACROS -D__STDC_CONSTANT_MACROS should fix this issue on the older compilers
__STDC_LIMIT_MACROS and __STDC_CONSTANT_MACROS are a workaround to allow C++ programs to use stdint.h macros specified in the C99 standard that aren't in the C++ standard.


andyli029:
I signed it!

PR #449, opened 2017-02-11
Prefer std::atomic over MemoryBarrier

sipa:
In port/atomic_pointer.h, the preference is to use MemoryBarrier if available, and only fall back to std::atomic if it isn't available.

I believe it would be preferable to use std::atomic, as it may allow better performance, and integrates without changes into ThreadSanitizer.

This PR changes the preference to std::atomic if available.

### DIFF
--- a/port/atomic_pointer.h
+++ b/port/atomic_pointer.h
@@ -46,6 +46,30 @@
 namespace leveldb {
 namespace port {
 
+// AtomicPointer based on <cstdatomic> if available
+#if defined(LEVELDB_ATOMIC_PRESENT)
+class AtomicPointer {
+ private:
+  std::atomic<void*> rep_;
+ public:
+  AtomicPointer() { }
+  explicit AtomicPointer(void* v) : rep_(v) { }
+  inline void* Acquire_Load() const {
+    return rep_.load(std::memory_order_acquire);
+  }
+  inline void Release_Store(void* v) {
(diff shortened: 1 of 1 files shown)


sipa:
I signed it!

droark:
tACK

Compiled and confirmed there were no errors. All tests passed, and the benchmarks (static and shared) seemed reasonable at a glance. Didn't benchmark vis-à-vis my fix (#422). sipa's fix, after thinking about it, seems preferable to mine, although I'll leave mine open in case the team thinks otherwise.

pwnall:
Thank you very much for your contribution!

The comment at the top of `atomic-pointer.h` states "Also, some <atomic> implementations are much slower than a memory-barrier based implementation (~16ns for <atomic> based acquire-load vs. ~1ns for a barrier based acquire-load)."

While I'm not sure this toolchain issue is relevant anymore, we'd need to do some testing to confirm. Do you happen to have access to the platforms that have MemoryBarrier implementations? If it turns out that std::atomic implementations in modern toolchains are good enough, we can remove the MemoryBarrier code altogether, because we'll require C++11 soon.

cmumford:
@googlebot rescan

cmumford:
@googlebot rescan

pwnall:
We removed AtomicPointer a while back. The codebase uses `std::atomic` directly now.

PR #450, opened 2017-02-22
Fixed sqlite3 benchmark to write to directory in --db flag

sporgj:
When running tests, it always wrote to the environment's default directory (/tmp/leveldb).
This change allows it to follow the behaviour exhibited by db/db_bench.cc


sporgj:
I signed it!

PR #456, opened 2017-03-09
Refactor db_impl file.

denleyhsiao:
Refactor Open function.

### DIFF
--- a/db/autocompact_test.cc
+++ b/db/autocompact_test.cc
@@ -24,7 +24,7 @@ class AutoCompactTest {
     DestroyDB(dbname_, options_);
     options_.create_if_missing = true;
     options_.compression = kNoCompression;
-    ASSERT_OK(DB::Open(options_, dbname_, &db_));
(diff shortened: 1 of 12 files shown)


denleyhsiao:
I signed it!

ghemawat:
Hi,

Some things I noticed.

1. I don't see what the change to add DBImpl::Open accomplishes.  Its code  being moved from one place to another, and the destination is a class private to the leveldb implementation so either way no code outside leveldb should care about where Open is implemented.

2. DB::Open in the future could be changed to instantiate a different class (e.g., if we add more engines and can dynamically pick between engines).

3. I don't understand the motivation for refactoring SanitizeOptions.

4. The third commit makes the util/ directory contain references to DBImpl, but the util directory is a lower-level directory than the db directory (db directory can depend on util, but util must not depend on db).

I suspect there is something you are trying to achieve with these rearrangements.  It would be good to know what that is so we can figure out what an appropriate solution is.  The current set of commits are probably not the right solution.

PR #459, opened 2017-03-20
Remove compiler warning in logging.cc

jonricha:
There is an unsigned vs signed comparison warning in logging.cc.  This should remove it.  It is already guaranteed to be a positive number from the condition on line 53 where c is already known to be between '0' and '9', so c will be in the range of 0 to 9.

### DIFF
--- a/util/logging.cc
+++ b/util/logging.cc
@@ -52,7 +52,7 @@ bool ConsumeDecimalNumber(Slice* in, uint64_t* val) {
     char c = (*in)[0];
     if (c >= '0' && c <= '9') {
       ++digits;
-      const int delta = (c - '0');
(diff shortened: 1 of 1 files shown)


jonricha:
I signed it

MarcoFalke:
Can be closed, since it is included in #492 and needs rebase.

pwnall:
@jonricha Thank you for your contribution!

I didn't merge it because the old implementation of `ConsumeDecimalNumber` crashed inexplicably on ARM until we changed the type of a local variable, so I was afraid that a similar type change might bring the crashes back. I ended up reimplementing `ConsumeDecimalNumber` and this PR doesn't make sense in the context of the new implementation.

Also, sorry for the late response.

PR #471, opened 2017-05-21
windows support added

padperso:
I have added a complete support for windows with win32 native api.
- Does compile on mingw or visual studio
- Does silll compile on linux :-)
- All tests passed ok on linux and windows
- Not fully posix (sorry), but should have no impact on this part of the code


solvingj:
Is this still under consideration?

pwnall:
We've just landed the official version, which is based on some code from Chromium. Thank you very much for your contribution, and sorry we couldn't get to this sooner!

Issue #478, opened 2017-06-15
Implement support for ARM64 crc32 instructions


zorrorffm:
ARM64 supports crc32 instructions which can use to accelerate crc32 computation.

cmumford:
This is currently implemented in the [crc32c library](https://github.com/google/crc32c) which will be integrated into the 1.21 release (see [announcement](https://groups.google.com/forum/#!topic/leveldb/GTDiWvgizMU)). Will leave open until implemented.

cmumford:
1.21 is released - closing.

PR #479, opened 2017-06-15
Optimize CRC32 calculation on arm64

zorrorffm:
ARM64 provides crc32 instructions for accelerating crc32 calculation.
This patch is optimization for linux under aarch64 and mainly refers
to src/common/crc32c_aarch64.c in ceph project.

The comparision of performance is as below

old
crc32c       :       5.670 micros/op;  688.9 MB/s (4K per op)

new
crc32c       :       0.327 micros/op; 11941.7 MB/s (4K per op)

Change-Id: Ia387e3b7f9f8e1fa11332b5be34ced4c4d5575ce

The name I signed CLA with is fangming.fang@arm.com
The issue number is #478

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -10,3 +10,4 @@ Sanjay Ghemawat <sanjay@google.com>
 # Partial list of contributors:
 Kevin Regan <kevin.d.regan@gmail.com>
 Johan Bilien <jobi@litl.com>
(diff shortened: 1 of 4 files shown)


zorrorffm:
These patch is not good enough. I will deliver new patch later.

PR #481, opened 2017-07-08
Remove unnecessary temporary variable 'k'

xiaosuo:


ghemawat:
My guess is that this change is not worth the amount of effort needed to sign the CLA, merge the commit, run the tests, etc..  The old code is clear enough as is.

xiaosuo:
@ghemawat It means that this micro cleanup is so obvious that no tests is needed. :)

cmumford:
@googlebot rescan

cmumford:
@xiaosuo I am going to side with sanjay and stick with the current implementation. Thank you for your pull request.

PR #488, opened 2017-07-21
Suggestion about AcceleratedCRC32C()

VincentVessalius:
I think the Process(unaligned bytes) in function in file port_posix_sse.cc maybe wrong. 
The original code does m times STEP1, where m = p%8.
If p=10,then m=2,and the new p will be equal to 12, which is not aligned by 8. 
Thannks for reading.

### DIFF
--- a/port/port_posix_sse.cc
+++ b/port/port_posix_sse.cc
@@ -93,7 +93,7 @@ uint32_t AcceleratedCRC32C(uint32_t crc, const char* buf, size_t size) {
 
   if (size > 16) {
     // Process unaligned bytes
-    for (unsigned int i = reinterpret_cast<uintptr_t>(p) % 8; i; --i) {
+    for (unsigned int i = reinterpret_cast<uintptr_t>(p) % 8; i && (i<8); i++) {
       STEP1;
     }
 


VincentVessalius:
I signed it!

pwnall:
Can you please submit a test that the original code fails, but your test passes? You can use the non-accelerated implementation to get a reference value.

VincentVessalius:
I know the original code can get the right result. However, the original code can’t do a right alignment work. 
  For example, when you use the original code, if _p_ = 0x0003, then reinterpret_cast<uintptr_t>(_p_) % 8 will be 3. That means you should do 5 times _STEP1_ so that _p_ will be added to 0x0008, which is aligned. But the original code will perform 3 times _STEP1_, which can just add _p_ to 0x0006.
  My suggestion will always let the _p_ aligned to a multiple of 8, which is the purpose of that code.
  Of course, the original code can get the right result because it just do a wrong alignment work, which makes the accelerate slow. As far as I know, the right alignment work can make the memory access faster.
  Thanks for reading. May you have a good summer!

pwnall:
@VincentVessalius Thank you for explaining! The explanation makes sense, and a fix is on the way.

pwnall:
This landed in 2964b803b857932ff7499d7bebb61dc5514dab7c. Thank you very much for your contribution!

PR #490, opened 2017-07-28
Optimize CRC32 calculation on arm64

zorrorffm:
ARM64 provides crc32 instructions for accelerating crc32 calculation.
This patch is optimization for linux under aarch64

The comparision of performance is as below

old
crc32c       :       5.670 micros/op;  688.9 MB/s (4K per op)

new
crc32c       :       0.451 micros/op; 8663.4.7 MB/s (4K per op)

Change-Id: I51d25ca19688fc95a57b84ee0c1493c18a288087

The name I signed CLA with is fangming.fang@arm.com
The issue number is #478

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -10,3 +10,4 @@ Sanjay Ghemawat <sanjay@google.com>
 # Partial list of contributors:
 Kevin Regan <kevin.d.regan@gmail.com>
 Johan Bilien <jobi@litl.com>
+Fangming Fang <Fangming.Fang@arm.com>
--- a/Makefile
+++ b/Makefile
@@ -422,3 +422,9 @@ $(STATIC_OUTDIR)/port/port_posix_sse.o: port/port_posix_sse.cc
 
 $(SHARED_OUTDIR)/port/port_posix_sse.o: port/port_posix_sse.cc
 	$(CXX) $(CXXFLAGS) $(PLATFORM_SHARED_CFLAGS) $(PLATFORM_SSEFLAGS) -c $< -o $@
+
+$(STATIC_OUTDIR)/port/port_posix_linux_arm64.o: port/port_posix_linux_arm64.cc
+	$(CXX) $(CXXFLAGS) $(PLATFORM_ARMV8_CRC32FLAGS) -c $< -o $@
+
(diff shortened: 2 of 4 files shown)


pwnall:
@zorrorffm Thank you very much for the contribution!

Due to some infrastructure issues, I won't be able to act on this very quickly. That being said, I am very grateful for your PR, and I look forward to improving the ARM performance.

zorrorffm:
@pwnall Thanks you.

If you need ARM machine to run some tests, I would be pleased to do that.

pwnall:
@zorrorffm I ran tests on a Google Pixel C tablet, and the optimized version seems to be 4-10x faster. If this is reasonably representative hardware, I don't think I'll need another machine. I'll take you up on your offer if it turns out I really need a machine that can easily run a GNU/Linux system.

pwnall:
@zorrorffm The LevelDB work hasn't happened yet, but I used your code in https://github.com/google/crc32c/blob/master/src/crc32c_arm64.cc -- thank you very much!

pwnall:
@zorrorffm We deployed the version of this patch at https://github.com/google/crc32c/blob/master/src/crc32c_arm64.cc in Chrome, and we got crashes when trying to execute the `vmull_p64` instruction on MSM8916 boards.

I have a hunch that we should also be checking that the flags returned by `hwcap` include `HWCAP_PMULL`, which is `(1 << 4)`, before enabling the AArch64 accelerated code. Is this correct?

pwnall:
@zorrorffm On the issue above, can you please take a look and comment on https://github.com/google/crc32c/pull/6?

zorrorffm:
@pwnall Yes, your fix is correct. Thank you for your findings.

zorrorffm:
Update the patch with detection of  CPU capabilities. Acceleration is available only if pmull and crc32c instructions are supported

cmumford:
@googlebot rescan

cmumford:
Thx for the PR @zorrorffm. The optimized crc32c code was moved out into https://github.com/google/crc32c (which supports armv8-a+crc+crypto) in 5c39524f3639e6bf6ab49215152d24273e662986.

PR #495, opened 2017-07-31
Remove a dangling pointer.

skilxn:


skilxn:
I signed it!

pwnall:
The variable you're assigning to isn't used anywhere in the code. Best case, the optimizer figures that out and removes the assignment as dead code. Worst case, we waste code and CPU cycles. 

Can you please explain the benefits you're hoping to get with this PR, to figure out if they're worth the costs?

pwnall:
Closing due to inactivity. Please answer the question above if you'd like this PR to be reconsidered.

PR #502, opened 2017-08-29
avoid lose efficacy for CompactRange

demiaowu:
After getting max_level_with_files value, there is no guarantee for each level state consistency. If some data which have been marking delete push to the higher level (eg, max_level_with_files+2, though it's hard to trigger) before the specify delete compaction, it will cause these data cannot be delete forever. Especially, in the multi-thread environment, If there is another thread call the DBImpl::CompactRange, eg. ceph. It will make this problem expose much more easy.

### DIFF
--- a/db/db_impl.cc
+++ b/db/db_impl.cc
@@ -579,6 +579,29 @@ void DBImpl::CompactRange(const Slice* begin, const Slice* end) {
   for (int level = 0; level < max_level_with_files; level++) {
     TEST_CompactRange(level, begin, end);
   }
+  
+  int old_max_level_with_file = max_level_with_files;
+  while (old_max_level_with_file < config::kNumLevels-1) {
+    {
+       MutexLock l(&mutex_);
+       Version *base = versions_->current();
+       for (int level = old_max_level_with_file + 1; level < config::kNumLevels; ++ level) {
+         if (base->OverlapInLevel(level, begin, end)) {
+           max_level_with_files = level;
+         }
+       }
+    }
+    
+    if (max_level_with_files != old_max_level_with_file) {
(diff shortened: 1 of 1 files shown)


demiaowu:
I signed it!

demiaowu:
Additions:
        For the function of `DBImpl::CompactRange` in `db/db_impl.cc`, it not a atomic operation. Although in the `TEST_CompactRange(level, begin, end)` , there is a condition variables used for keeping synchronization. But it is the guarantee of `happen-before`  order, not atomicity. Thus, the Minor compaction and Size Compaction will take part in the competition. After getting max_level_with_files value, the mutex of  `MutexLock`  will be unlock, all of them (Manual compation, minor compation and size compation ) have opportunity to be executed. Thus, there are some data  which have been marking delete might be pushed to the higher level before the specify delete compaction. If the `level >= max_level_with_files+1`,  these data cannot be delete in this `CompactRange`. 
567   void DBImpl::CompactRange(const Slice* begin, const Slice* end) { 
568   int max_level_with_files = 1; 
569   { 
570     MutexLock l(&mutex_); 
571     Version* base = versions_->current(); 
572     for (int level = 1; level < config::kNumLevels; level++) { 
573       if (base->OverlapInLevel(level, begin, end)) { 
574         max_level_with_files = level; 
575       } 
576     } 
577   } 
578   TEST_CompactMemTable(); // TODO(sanjay): Skip if memtable does not overlap 
579   for (int level = 0; level < max_level_with_files; level++) { 
580     TEST_CompactRange(level, begin, end); 
581   } 
582 }

cmumford:
@googlebot rescan

PR #538, opened 2018-01-05
Update table.h

dcorderoch:
Fix typo

### DIFF
--- a/include/leveldb/table.h
+++ b/include/leveldb/table.h
@@ -24,7 +24,7 @@ class TableCache;
 // multiple threads without external synchronization.
 class LEVELDB_EXPORT Table {
  public:
-  // Attempt to open the table that is stored in bytes [0..file_size)
+  // Attempt to open the table that is stored in bytes [0..file_size]
   // of "file", and read the metadata entries necessary to allow
   // retrieving data from the table.
   //


dcorderoch:
I signed it!

include/leveldb/table.h
```diff
@@ -24,7 +24,7 @@ class TableCache;
 // multiple threads without external synchronization.
 class LEVELDB_EXPORT Table {
  public:
-  // Attempt to open the table that is stored in bytes [0..file_size)
+  // Attempt to open the table that is stored in bytes [0..file_size]
```
ghemawat: Actually, the old use of ")" was intentional.  In Math syntax for intervals,
   [a..b)
means a half-closed interval that included b-1, but not b.

So the old comment was conveying that Table contents are stored in the bytes numbered 0 through file_size-1 of the file (i.e., the first file_size bytes).

dcorderoch: oh, I actually learned that as [a,b[ for a set that goes from a, and up to b-1, but not b, didn't know about that notation, but I just checked, and apparently, the way I learned is the non-standard notation

then given that's not really a typo, should I just close this PR?

ghemawat:
Interesting.  I was unaware of the [a,b[ notation until now.

I am closing the PR.

PR #543, opened 2018-01-15
Update version_set.h

langsiji:
fix comment


langsiji:
I signed it!

langsiji:
I signed it!

pwnall:
"iff" has more meaning than "if"... it means "if and only if". See https://en.wikipedia.org/wiki/If_and_only_if

PR #551, opened 2018-02-04
Fixed style Names and Order of Includes.

cppcoffee:
use http://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes style.

### DIFF
--- a/helpers/memenv/memenv.cc
+++ b/helpers/memenv/memenv.cc
@@ -4,14 +4,14 @@
 
 #include "helpers/memenv/memenv.h"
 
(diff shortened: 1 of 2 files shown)


cmumford:
Thx for the PR @cppcoffee. The headers in helpers/memenv/memenv.cc were corrected in another commit. We are in the process of running clang-format on all leveldb files (which will fix memenv_test.cc) as well as many many other style issues. My preference is to land all formatting changes (if possible) in a single commit to keep "git blame" as useful as possible.

PR #552, opened 2018-02-06
Persistent Snapshots

iljaiwas:
This PR adds to ability to persist snapshots across application sessions.

This feature can be very useful, if you have a long running operations (e.g. syncing a multi-GB database to the cloud), but the user has the ability to kill your app.


DaMatrix:
not sure what the status is on this, but it would be interesting to see this added at some point

pwnall:
Sorry, the concept is interesting indeed, but I don't know of any internal need for this.

PR #553, opened 2018-02-08
Implement AtomicPointer for s390

WandyLau:
Signed-off-by: Wang Dong <xdragon007@gmail.com>

### DIFF
--- a/port/atomic_pointer.h
+++ b/port/atomic_pointer.h
@@ -41,6 +41,8 @@
 #define ARCH_CPU_PPC_FAMILY 1
 #elif defined(__mips__)
 #define ARCH_CPU_MIPS_FAMILY 1
+#elif defined(__s390__) || defined(__s390x__)
+#define ARCH_CPU_S390_FAMILY 1
 #endif
 
 namespace leveldb {
@@ -116,6 +118,12 @@ inline void MemoryBarrier() {
 #elif defined(ARCH_CPU_MIPS_FAMILY) && defined(__GNUC__)
 inline void MemoryBarrier() {
   __asm__ __volatile__("sync" : : : "memory");
+}
+#define LEVELDB_HAVE_MEMORY_BARRIER
+
+#elif defined(ARCH_CPU_S390_FAMILY) && defined(__linux__)
+inline void MemoryBarrier() {
+  __asm__ __volatile__("" : : : "memory");
 }
 #define LEVELDB_HAVE_MEMORY_BARRIER
 
@@ -235,6 +243,7 @@ class AtomicPointer {
 #undef ARCH_CPU_ARM_FAMILY
 #undef ARCH_CPU_ARM64_FAMILY
 #undef ARCH_CPU_PPC_FAMILY
+#undef ARCH_CPU_S390_FAMILY
 
 }  // namespace port
 }  // namespace leveldb


WandyLau:
Hi, maintainer

I have signed the CLA. Would you please merge this patch?

pwnall:
@WandyLau Thank you very much for your contribution!

Can you please briefly explain (here or in a comment in code) why the `__asm__ __volatile__("" : : : "memory")` idiom is safe for S390?

Linux seems to use `bcr 15,0` to get a memory barrier on S390. -- https://github.com/torvalds/linux/blob/master/arch/s390/include/asm/barrier.h

At the same time, Linux uses `mfence` for barriers on X86 -- https://github.com/torvalds/linux/blob/master/arch/x86/include/asm/barrier.h

We don't seem to need `mfence` because the memory model guarantees sequential consistency, so I imagine there's an explanation for S390 as well.

Thank you very much!

pwnall:
Skimming through http://ieeexplore.ieee.org/stamp/stamp.jsp?arnumber=508100 suggests that S390 also guarantees sequential consistency (using a directory to implement cache coherence) just like X86. I'd like some confirmation that I understood this correctly, or the correct explanation.

Thank you!

mitch081214:
1. - # Mitmm081214@gmail.com _~~is ~~ _ ___ I don't __ __ I don't think you know what real is tell it beats you over the head with s o f and it says it's from you ___ I don't know what real is tell it beats you over the head with s o f and it says it's from you __ __ I don't know if I'm coming home now I could pull the certificate and crash the whole world wide web _____ I __ __ I don't think you know what real is tell it beats you over the head with s o f and it says it's __ __ you __ _ 

_Gesendet von meinem Samsung SM-G930T mit [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

WandyLau:
@pwnall Yes, as I discussed with some guy in the s390 team, it is like x86. And this implementation is a standard one as intel(x86).  I don't think this will have something to do with mfence just like  your previous response.

WandyLau:
@pwnall Hi, I wonder if this is good to merge?

pwnall:
@WandyLau Sorry for taking so long to get back to you! 

It's neat that you have access to the S390 team! Can you please point to some public document that describes the S390 memory consistency model from a technical angle (the one I found has a heavy marketing slant) and makes it clear that your implementation is correct?

On a very related note, LevelDB will require C++11 support by the next release. If your compilers don't have that support, you won't be able to build future version of LevelDB.

WandyLau:
@pwnall 
I am not sure if there is any open document related to this.
I googled some, but without any result.
I have a mail for the guy once I got the answer I will tell you. He also said as below:

**this seems to be a standard acquire / release barrier  
if you have GCC support for atomic, this should already work with the code as is  
or else, you can implement the "memory barrier" variant, using the same compiler-only that is used on Intel:  

asm volatile("" : : : "memory");** 

Yes, we will update the gcc accordingly.

pwnall:
@WandyLau I've landed the changes I've mentioned in this thread. I think that you shouldn't need this PR, unless the `<atomic>` implementation in the S390 toolchain is worse than the memory barrier listed here.

Also, in the not-too-long term, we plan to remove the special cases in this file and assume that toolchains implement `<atomic>` sensibly. Would you still like to discuss this PR, or can it be closed?

WandyLau:
@pwnall What do you mean by "landed"? I have asked the guy, but no response so far. I will ask him later again. You said you will remove special cases, I am not sure what do you mean. I hope this patch
can be merged to support s390. No discussion I think. If the proof of the this work is required, I will ask the guy later soon.

pwnall:
@WandyLau The changes I mentioned have been developed in an internal repository, and were committed to the open source repository recently. This is what I meant by "landed". The master branch of LevelDB currently requires a compiler with C++11 support, and is built using CMake.

If your have a C++11 toolchain, you should have the `<atomic>` header, and AtomicPointer can take advantage of that. You should never get the `Please implement AtomicPointer for this platform.` error. On the flip side, if you don't have a C++11 toolchain, you won't be able to build LevelDB.

Can you please try building the master branch of LevelDB on your system, and consider whether this PR is still needed?

WandyLau:
yes, that works. I think this can be closed then. Thanks.

pwnall:
@WandyLau Thank you very much!

kycfeel:
Hey guys. I'm trying to build the leveldb on IBM `s390x` (not the `s390`).

I'm getting the `error: #error Please implement AtomicPointer for this platform.` error even after I get the C++11 toolchain for the g++ (`alias g++="g++ --std=c++11"`). 

Can someone give me an advise?

Thank you.

PR #558, opened 2018-02-27
Fix TargetFileSize() mixed unsigned/signed warning

eklitzke:
This fixes a minor compiler warning when building with GCC 7.3. This is the only warning when building the project itself, there are some other warnings when building the test cases that I didn't look into.


MarcoFalke:
See #492

eklitzke:
Makes sense, I'll close this.

PR #563, opened 2018-03-14
Can “DeleteObsoleteFiles” not be locked?

likuoping136:
When "is_manual = true",this function "DeleteObsoleteFiles()" may be has so many file to del,so "compact" will be locked long time. and then "leveldb" can't be write。


likuoping136:
I signed it!

PR #568, opened 2018-03-20
Fix cmake's installation, which does not install 'leveldb/export.h'

Hyzeta:
I build leveldb using **cmake** on my Mac, then I tried to link leveldb with my example, but the cmake tells me that:

```
/Users/hyzeta/Storage/Projects/domaindb/leveldb/include/leveldb/db.h:10:10: fatal error: 'leveldb/export.h' file not found
#include "leveldb/export.h"
         ^~~~~~~~~~~~~~~~~~
1 error generated.
```

I findout that the origin CMakeLists.txt  does **not** install 'leveldb/export.h' file. So I modify the origin CMakeLists.txt and add 'leveldb/export.h' file while leveldb's installation.

Then I build the leveldb with new CMakeLists.txt (There is no wrong while building and no wrong while testing), and link leveldb with my example. Error nolonger exists.


pwnall:
Whoops, that's correct. Thanks!

pwnall:
I'll close this PR when the relevant commit lands from our internal tree.

Issue #570, opened 2018-03-22
Carthage Support.


uny:
I'm trying to make [firebase-ios-sdk](https://github.com/firebase/firebase-ios-sdk) [carthage](https://github.com/Carthage/Carthage) supported as dynamic framework.
https://github.com/firebase/firebase-ios-sdk/issues/9
To do that, the dependencies are also needed to be carthage supported.

pwnall:
This repository supports a single build system -- CMake. This is all we have resources for right now.

Out of the box, CMake gives us support for most desktop Unixes, Mac, and Android. We know it will support Windows as well. We hope CMake also has some form of iOS support, but haven't looked into that yet. This matters because it means CMake supports a lot of platforms with its configuration file.

I don't know any build system that would pull its weight as a second supported system -- it would have to add support for so many important platforms that we'd consider it's worth the extra maintenance burden.

If you have strong arguments for why Carthage is that system, please explain here. I am closing the feature request because we don't intend to implement or accept a PR for it, but I will still see your comments.

PR #571, opened 2018-03-22
Carthage Support.

uny:
https://github.com/google/leveldb/issues/570


pwnall:
I'm sorry, but I don't want to maintain this configuration. It's quite a bit of code, and there's zero continuous integration support of it, so I'd have to learn how to test it manually, and do the tests for every PR / commit.

Can you accomplish your goals in another way? For example, you could set up a repository that includes google/leveldb as a submodule and wraps your configuration around it.

PR #590, opened 2018-05-21
Updated CopyString function

mowais786:
Update CopyString function to return a NULL terminated string. I received garbage strings while using the C interface in Swift.were

### DIFF
--- a/db/c.cc
+++ b/db/c.cc
@@ -143,7 +143,8 @@ static bool SaveError(char** errptr, const Status& s) {
 }
 
 static char* CopyString(const std::string& str) {
-  char* result = reinterpret_cast<char*>(malloc(sizeof(char) * str.size()));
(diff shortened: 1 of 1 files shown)


mowais786:
I signed it!

ghemawat:
CopyString is used for values returned by leveldb_get.  These values can be binary data, not just text and can therefore contain embedded null characters.  Therefore we do not rely on null-termination to indicate the length of the string, but instead return the length (e.g., leveldb_get's vallen argument).

We could in addition also null-terminate for the convenience of people who are storing text, but I would rather not do that because it would encourage libraries to start relying on that and those libraries would then break in a very bad way (silent truncation) if the database ever contained a value with embedded nulls.

I suggest that if you are only dealing with plain-text values, you write a simple wrapper around leveldb_get in your application.

PR #595, opened 2018-06-09
add class Env implement for Windows OS in env_posix.cc

yanrk:
add class Env implement for Windows OS in env_posix.cc


cmumford:
Thanks for the pull request @yanrk. Although it landed after your pull request, leveldb added native Windows support in c69d33b0ec3dad2a8063ad66da9d51a1d6309f4e. If you have an opportunity please give it a try.

yanrk:
ok, the new version does well now

PR #599, opened 2018-06-18
update version_set.cc

RCmerci:
add `GetIncludedInputs` function, which collect files included in [begin,end], unlike `GetOverlappingInputs`.

`SetupOtherInputs` calls `GetIncludedInputs` to make `expanded0` included in [all_start, all_limit]  .

it makes more likely to expand level+1


RCmerci:
I signed it!

PR #600, opened 2018-06-18
GetOverlappingInputs:decrease loop time

RCmerci:
handle file_start, file_limit 2 cases in one loop


cmumford:
@googlebot rescan

PR #606, opened 2018-07-18
Use of CMAKE_INSTALL_INCLUDEDIR before inclusion of GNUInstallDirs.cmake gives bogus leveldb import target

pwidene:
CMAKE_INSTALL_INCLUDEDIR was being used to set a leveldb target property for the INSTALL_INTERFACE. This variable is defined in the GNUInstallDirs module, which hadn't been included at that point. Since the variable was empty, no INTERFACE_INCLUDE_DIRECTORIES property was defined on the installed leveldb::leveldb import target, making it hard to use that target properly in importing CMake setups.


pwidene:
Signed.

PR #613, opened 2018-08-26
Increase maximum read-only mmap()s used from 1000 to 4096 on 64-bit systems

luke-jr:
By default LevelDB will only mmap() up to 1000 ldb files for reading and then fall back
to using file desciptors.

The typical linux system has a 'vm.max_map_count = 65530', so mapping only 1000 files
seems arbitarily small. Increase this value to another arbitrarily small value, 4096.

See also: https://github.com/bitcoin-core/leveldb/pull/19


pwnall:
Please read the contributing instructions before submitting PRs.

PR #615, opened 2018-08-27
Add API to create unique environments, with custom limits

luke-jr:
Instead of simply increasing a hard-coded internal mmap limit (#613), allow the software using LevelDB to choose its own limits with a new `EnvOptions` struct.

This makes the `env_posix_test` cleaner, since it was using an internal hack to override the limits already.

Considerations:

* Up until now, `PosixEnv` was never destroyed, and destroying it aborted. Now that it can be destroyed, does it clean up properly? (No obvious problems jumped out at me.)
* This is the first time `PosixEnv` can have multiple instances. Is that safe? (Again, no obvious issues jump out.)
* `EnvOptions` as currently designed would need to be extended to support any arbitrary environment options desired. Since there is only one supported platform for now, that's probably not an issue, but would it make more sense to have a `std::map<std::string, std::string>` for flexibility here?


pwnall:
You can pass a custom Env to https://github.com/google/leveldb/blob/master/include/leveldb/options.h#L63

PosixEnv is not part of the API by design. We don't want to be constrained by any of the implementation details in there. If your application has tight performance needs, you're better off forking PosixEnv and adding your tweaks on top.

PR #648, opened 2018-12-23
Fix cmake configuration include not exported.

Invincibl-e:
reference:[creating-relocatable-packages](https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#creating-relocatable-packages)


Invincibl-e:
The https://github.com/google/leveldb/pull/606 solution is better than me, I close this pull request.

PR #652, opened 2019-01-22
VersionEdit::Clear() should also clear compact_pointers_

caodhuan:


pwnall (approved):
LGTM. Thank you for the fix!

PR #681, opened 2019-04-29
 CMake: Add support for Hunter package manager dependencies 

chfast:
[Hunter](http://hunter.sh) is a package manager for CMake based packages. This changes would allow using the upstream leveldb in Hunter. With these changes, when Hunter is enabled, the dependencies are treated as Hunter packages.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,14 @@
 # found in the LICENSE file. See the AUTHORS file for names of contributors.
 
 cmake_minimum_required(VERSION 3.9)
+
+option(HUNTER_ENABLED "Enable Hunter package manager" OFF)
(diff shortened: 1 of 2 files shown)


pwnall:
@chfast Thank you very much for your contributions around Google repositories!

I'd really rather not get into the business of supporting individual package managers. I was hoping that our CMake build is flexible enough that package managers can wrap around us easily. Are there things we could do that aren't specific to one package manager?

chfast:
Unfortunately, there is no way to handle this in Hunter without these changes. But I understand the decision.
I will sill send some generic CMake improvements later on that will help keeping the Hunter related changes minimal.

The Hunter package is maintained in https://github.com/hunter-packages/leveldb.

Issue #685, opened 2019-05-07
CMake: Use find_package() to find snappy


chfast:
Create proper FindSnappy.cmake module and use `find_package()` to find the dependency. Create an imported target representing the library.

Also consider using `find_package(... CONFIG)` as snappy exports CMake package config files.

pwnall:
Thank you for your contribution!

We're not currently interested in contributions to the build configuration. We now have a paragraph explaining the situation in  https://github.com/google/leveldb/blob/master/README.md#contributing-to-the-leveldb-project

PR #686, opened 2019-05-08
 cmake: Use find_package() to find Snappy 

chfast:


olzhabay:
#685

pwnall:

We're not currently interested in contributions to the build configuration. We now have a paragraph explaining the situation in  https://github.com/google/leveldb/blob/master/README.md#contributing-to-the-leveldb-project

PR #698, opened 2019-06-11
Use Version::ForEachOverlapping to implement Version::Get()

neal-zhu:
This code has been tested with existing tests and no new tests are written because it justs implements  Version:: Get() with Version:: ForEachOverlapping() which has been fully tested.

### DIFF
--- a/db/version_set.cc
+++ b/db/version_set.cc
@@ -332,9 +332,6 @@ Status Version::Get(const ReadOptions& options, const LookupKey& k,
     GetStats* stats;
     const ReadOptions* options;
     Slice ikey;
-    Slice user_key;
-    const Comparator* ucmp;
-    std::string* value;
     FileMetaData* last_file_read;
     int last_file_read_level;
 
@@ -364,14 +361,15 @@ Status Version::Get(const ReadOptions& options, const LookupKey& k,
       }
       switch (state->saver.state) {
         case kNotFound:
-          return true;  // Keep saerching in other files
+          return true;  // Keep searching in other files
         case kFound:
           state->found = true;
           return false;
         case kDeleted:
           return false;
         case kCorrupt:
-          state->s = Status::Corruption("corrupted key for ", state->user_key);
+          state->s =
+              Status::Corruption("corrupted key for ", state->saver.user_key);
           state->found = true;
           return false;
       }
@@ -386,17 +384,14 @@ Status Version::Get(const ReadOptions& options, const LookupKey& k,
 
   state.options = &options;
   state.ikey = k.internal_key();
-  state.user_key = k.user_key();
-  state.ucmp = vset_->icmp_.user_comparator();
-  state.value = value;
   state.vset = vset_;
 
   state.saver.state = kNotFound;
-  state.saver.ucmp = state.ucmp;
-  state.saver.user_key = state.user_key;
-  state.saver.value = state.value;
+  state.saver.ucmp = vset_->icmp_.user_comparator();
+  state.saver.user_key = k.user_key();
+  state.saver.value = value;
 
-  ForEachOverlapping(state.user_key, state.ikey, &state, &State::Match);
+  ForEachOverlapping(state.saver.user_key, state.ikey, &state, &State::Match);
 
   return state.found ? state.s : Status::NotFound(Slice());
 }
--- a/db/version_set.cc
+++ b/db/version_set.cc
@@ -327,7 +327,6 @@ Status Version::Get(const ReadOptions& options, const LookupKey& k,
   Slice ikey = k.internal_key();
   Slice user_key = k.user_key();
   const Comparator* ucmp = vset_->icmp_.user_comparator();
-
   stats->seek_file = nullptr;
   stats->seek_file_level = -1;
 
@@ -339,7 +338,7 @@ Status Version::Get(const ReadOptions& options, const LookupKey& k,
     const Comparator* ucmp;
     std::string* value;
     FileMetaData* last_file_read;
-    int last_file_level;
+    int last_file_read_level;
 
     VersionSet* vset;
     Status s;
@@ -347,14 +346,15 @@ Status Version::Get(const ReadOptions& options, const LookupKey& k,
     static bool Match(void* arg, int level, FileMetaData* f) {
       State* state = reinterpret_cast<State*>(arg);
 
-      if (state->last_file_read != nullptr &&
-          state->stats->seek_file == nullptr) {
+      if (state->stats->seek_file == nullptr &&
+          state->last_file_read != nullptr) {
         // We have had more than one seek for this read.  Charge the 1st file.
         state->stats->seek_file = state->last_file_read;
-        state->stats->seek_file_level = state->last_file_level;
+        state->stats->seek_file_level = state->last_file_read_level;
       }
+
       state->last_file_read = f;
-      state->last_file_level = level;
+      state->last_file_read_level = level;
 
       Saver saver;
       saver.state = kNotFound;
@@ -388,8 +388,9 @@ Status Version::Get(const ReadOptions& options, const LookupKey& k,
   state.s = Status::NotFound(Slice());
   state.stats = stats;
   state.last_file_read = nullptr;
-  state.last_file_level = -1;
+  state.last_file_read_level = -1;
 
+  state.options = &options;
   state.ikey = ikey;
   state.user_key = user_key;
   state.ucmp = ucmp;
--- a/db/version_set.cc
+++ b/db/version_set.cc
@@ -281,7 +281,6 @@ static bool NewestFirst(FileMetaData* a, FileMetaData* b) {
 
 void Version::ForEachOverlapping(Slice user_key, Slice internal_key, void* arg,
                                  bool (*func)(void*, int, FileMetaData*)) {
-  // TODO(sanjay): Change Version::Get() to use this function.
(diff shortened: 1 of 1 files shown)


neal-zhu:
I signed it!

neal-zhu:
I'm really confused, how come tests like memenv_test failed in [this build](https://travis-ci.org/google/leveldb/jobs/544205038) while this commit has nothing to do with memenv?

db/version_set.cc
```diff
@@ -330,94 +330,69 @@ Status Version::Get(const ReadOptions& options, const LookupKey& k,
...
       }
     }
-  }
+  };
+
+  stats->seek_file = nullptr;
+  stats->seek_file_level = -1;
+
+  State state;
+  state.s = Status::NotFound(Slice());
```
ghemawat: Non-OK status values are somewhat expensive.  I suggest adding a "bool found" to State, and if after everything is done without an error and found is false, create the NotFound status at that point.

db/version_set.cc
```diff
@@ -330,94 +330,69 @@ Status Version::Get(const ReadOptions& options, const LookupKey& k,
...
+  state.stats = stats;
+  state.ikey = ikey;
+  state.user_key = user_key;
+  state.ucmp = ucmp;
+  state.value = value;
+  state.vset = vset_;
+
+  ForEachOverlapping(user_key, ikey, &state, &State::Match);
```
ghemawat: Get() can be performance sensitive.  Have you run db_bench with full optimizations on before and after this change?

db/version_set.cc
```diff
@@ -330,94 +330,69 @@ Status Version::Get(const ReadOptions& options, const LookupKey& k,
...
-          // All of "tmp2" is past any data for user_key
-          files = nullptr;
-          num_files = 0;
-        } else {
-          files = &tmp2;
-          num_files = 1;
-        }
-      }
-    }
+    static bool Match(void* arg, int level, FileMetaData* f) {
```
ghemawat: I am not sure how portable these nested static functions are.  Can you hoist Match out so it is no longer nested (and you will also have to move State out)?  If you do that, they should probably be renamed to something like:

State => GetState
Match => GetMatcher

db/version_set.cc
```diff
@@ -330,94 +330,69 @@ Status Version::Get(const ReadOptions& options, const LookupKey& k,
...
+      State* state = reinterpret_cast<State*>(arg);
 
-    for (uint32_t i = 0; i < num_files; ++i) {
-      if (last_file_read != nullptr && stats->seek_file == nullptr) {
+      if (state->stats->seek_file == nullptr) {
```
ghemawat: This looks like a change in semantics, and might lead to performance problems, bad stats.

Suppose just a single file is consulted: the old code would not set seek_file, but the new code will set seek_file to the file that was consulted.

Any particular reason not to track last_file_read like the old code was doing?

db/version_set.cc
```diff
@@ -330,94 +330,69 @@ Status Version::Get(const ReadOptions& options, const LookupKey& k,
...
-  }
+  };
+
+  stats->seek_file = nullptr;
+  stats->seek_file_level = -1;
+
+  State state;
+  state.s = Status::NotFound(Slice());
+  state.stats = stats;
+  state.ikey = ikey;
```
ghemawat: Any particular reason to have variables like ikey any more?  You could instead initialize state.ikey directly when the function starts.  Same for user_key, ucmp.

Also, is the "Status s;" variable declared near the top of Get() still needed?

db/version_set.cc
neal-zhu: I intended to implement it with a class in an anonymous namespace until I saw the code in [Version::RecordReadSample](https://github.com/google/leveldb/blob/master/db/version_set.cc#L442).

db/version_set.cc
neal-zhu: Before:
LevelDB:    version 1.22
Keys:       16 bytes each
Entries:    1000000
RawSize:    110.6 MB (estimated)
FileSize:   62.9 MB (estimated)
fillseq      :       8.881 micros/op;   12.5 MB/s
fillsync     :    3992.566 micros/op;    0.0 MB/s (1000 ops)
fillrandom   :      11.409 micros/op;    9.7 MB/s
overwrite    :      12.871 micros/op;    8.6 MB/s
readrandom   :       6.165 micros/op; (1000000 of 1000000 found)
readrandom   :       5.795 micros/op; (1000000 of 1000000 found)
readseq      :       0.261 micros/op;  423.9 MB/s
readreverse  :       0.467 micros/op;  236.9 MB/s
compact      : 1089242.000 micros/op;
readrandom   :       4.761 micros/op; (1000000 of 1000000 found)
readseq      :       0.311 micros/op;  355.3 MB/s
readreverse  :       0.475 micros/op;  232.8 MB/s
fill100K     :    1422.327 micros/op;   67.1 MB/s (1000 ops)
crc32c       :       1.551 micros/op; 2519.0 MB/s (4K per op)

After
LevelDB:    version 1.22
Keys:       16 bytes each
Entries:    1000000
RawSize:    110.6 MB (estimated)
FileSize:   62.9 MB (estimated)
fillseq      :       8.603 micros/op;   12.9 MB/s
fillsync     :    3915.398 micros/op;    0.0 MB/s (1000 ops)
fillrandom   :      11.243 micros/op;    9.8 MB/s
overwrite    :      12.335 micros/op;    9.0 MB/s
readrandom   :       6.263 micros/op; (1000000 of 1000000 found)
readrandom   :       5.693 micros/op; (1000000 of 1000000 found)
readseq      :       0.236 micros/op;  468.9 MB/s
readreverse  :       0.399 micros/op;  277.2 MB/s
compact      : 1011847.000 micros/op;
readrandom   :       4.582 micros/op; (1000000 of 1000000 found)
readseq      :       0.203 micros/op;  544.6 MB/s
readreverse  :       0.309 micros/op;  358.4 MB/s
fill100K     :    1386.870 micros/op;   68.8 MB/s (1000 ops)
crc32c       :       1.349 micros/op; 2895.0 MB/s (4K per op)
snappycomp   :       3.847 micros/op; 1015.5 MB/s (output: 55.1%)
snappyuncomp :       0.613 micros/op; 6372.0 MB/s
snappycomp   :       4.281 micros/op;  912.4 MB/s (output: 55.1%)
snappyuncomp :       0.645 micros/op; 6056.2 MB/s

db/version_set.cc
neal-zhu: You are right, there is no need to have them anymore.

db/version_set.cc
neal-zhu: I cache the Saver object in State to avoid creating Saver for each overlapping file(though it may cause extra expense if there is no overlapping files) and the bench result is:
fillseq      :       8.472 micros/op;   13.1 MB/s
fillsync     :    3992.071 micros/op;    0.0 MB/s (1000 ops)
fillrandom   :      11.059 micros/op;   10.0 MB/s
overwrite    :      12.333 micros/op;    9.0 MB/s
readrandom   :       5.943 micros/op; (1000000 of 1000000 found)
readrandom   :       5.396 micros/op; (1000000 of 1000000 found)
readseq      :       0.219 micros/op;  505.2 MB/s
readreverse  :       0.350 micros/op;  315.8 MB/s
compact      :  871343.000 micros/op;
readrandom   :       4.140 micros/op; (1000000 of 1000000 found)
readseq      :       0.177 micros/op;  625.3 MB/s
readreverse  :       0.310 micros/op;  357.2 MB/s
fill100K     :    1347.276 micros/op;   70.8 MB/s (1000 ops)
crc32c       :       1.395 micros/op; 2799.8 MB/s (4K per op)
snappycomp   :       4.086 micros/op;  956.0 MB/s (output: 55.1%)
snappyuncomp :       0.628 micros/op; 6222.3 MB/s

db/version_set.cc
```diff
@@ -356,50 +355,50 @@ Status Version::Get(const ReadOptions& options, const LookupKey& k,
...
   state.last_file_read = nullptr;
   state.last_file_read_level = -1;
 
   state.options = &options;
-  state.ikey = ikey;
-  state.user_key = user_key;
-  state.ucmp = ucmp;
+  state.ikey = k.internal_key();
+  state.user_key = k.user_key();
+  state.ucmp = vset_->icmp_.user_comparator();
```
ghemawat: It looks like the following data is being placed both in state and state.saver.  I don't think that is necessary.  Can you drop the state fields that are duplicates of fields in Saver?

ucmp
user_key

db/version_set.cc
```diff
@@ -356,50 +355,50 @@ Status Version::Get(const ReadOptions& options, const LookupKey& k,
...
-        state->s = s;
+                                                &state->saver, SaveValue);
+      if (!state->s.ok()) {
+        state->found = true;
         return false;
       }
-      switch (saver.state) {
+      switch (state->saver.state) {
         case kNotFound:
           return true;  // Keep saerching in other files
```
ghemawat: Change "saerching" to "searching"

db/version_set.cc
neal-zhu: Right on it, maybe also drop the value field in State?

neal-zhu:
@googlebot I fixed it.

neal-zhu:
@ghemawat sorry for delaying for so long!

neal-zhu:
THANKS !!!!! This means a lot to me!!!

PR #702, opened 2019-07-06
add disableWAL option

kylezh:
When WAL is disabled, performance of 'fillseq' is improved a lot.

When WAL is disabled, performance of 'fillseq' is improved a lot.

This option is helpful when doing bulk load. And in-memory data will
be flushed to disk if CompactRange(nullptr, nullptr) is called at the
end of load.

```
$ time ./db_bench --benchmarks=fillseq --db=/tmp/db --disable_wal=0
LevelDB:    version 1.22
Date:       Sat Jul  6 23:51:41 2019
CPU:        16 * Intel(R) Core(TM)2 Duo CPU     T7700  @ 2.40GHz
CPUCache:   4096 KB
Keys:       16 bytes each
Entries:    1000000
RawSize:    110.6 MB (estimated)
FileSize:   62.9 MB (estimated)
fillseq      :       4.783 micros/op;   23.1 MB/s

real    0m4.798s
user    0m3.708s
sys     0m2.005s

$ time ./db_bench --benchmarks=fillseq --db=/tmp/db --disable_wal=1
LevelDB:    version 1.22
Date:       Sat Jul  6 23:52:02 2019
CPU:        16 * Intel(R) Core(TM)2 Duo CPU     T7700  @ 2.40GHz
CPUCache:   4096 KB
Keys:       16 bytes each
Entries:    1000000
RawSize:    110.6 MB (estimated)
FileSize:   62.9 MB (estimated)
fillseq      :       1.227 micros/op;   90.2 MB/s

real    0m1.253s
user    0m1.657s
sys     0m0.233s
```

### DIFF
--- a/benchmarks/db_bench.cc
+++ b/benchmarks/db_bench.cc
@@ -111,6 +111,9 @@ static bool FLAGS_reuse_logs = false;
 // Use the db with the following name.
 static const char* FLAGS_db = nullptr;
 
+// If true, disable wal  when writing.
+static bool FLAGS_disable_wal = false;
+
 namespace leveldb {
 
 namespace {
@@ -418,6 +421,9 @@ class Benchmark {
   }
 
   ~Benchmark() {
+    if (FLAGS_disable_wal) {
+      db_->CompactRange(nullptr, nullptr);
+    }
     delete db_;
     delete cache_;
     delete filter_policy_;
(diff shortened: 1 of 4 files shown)


ghemawat:
This is a very unsafe thing to turn on, and I would not like it to be so easy for people to make mistakes by turning it on without understanding the implications.  If you really care, you can use a custom Env where you make the files uses for the WAL return without writing anything.

PR #724, opened 2019-09-11
update .gitignore

kylezh:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,8 @@
 # Build directory.
 build/
 out/
+
+*.a
+*.o
+*.so
+*.so.*


pwnall:
What is the benefit of this CL?

Asking because the build process documented in the README will place all output in a directory, and we already have the most popular directory names covered.

kylezh:
@pwnall You are right. Because the upstream has switched to CMake, so for most of the case, all these files would be put in the build directory. So there is no benefit.

I met this problem because I always switched between upstream and old versions which are using Makefile. In this case, ignoring these files would be helpful. But this should not be a common case.

PR #730, opened 2019-09-14
Fix dismatch of code and comment

orrorcol:
The code comment didn't match the code:
https://github.com/google/leveldb/blob/21304d41f77990b8edabbdab33b222bd5ceb5f18/db/version_set.h#L62-L75

I fixed it.


krishnacx:
Would it be mismatch instead of dismatch in commit message ?

PR #737, opened 2019-09-27
Add vcpkg installation instructions

grdowns:
LevelDB is available as a port in [vcpkg](https://github.com/Microsoft/vcpkg), a C++ library manager that simplifies installation for LevelDB and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build LevelDB, ready to be included in their projects. 

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP, ARM) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/leveldb/portfile.cmake). We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,18 @@ Authors: Sanjay Ghemawat (sanjay@google.com) and Jeff Dean (jeff@google.com)
   * Only a single process (possibly multi-threaded) can access a particular database at a time.
   * There is no client-server support builtin to the library.  An application that needs such support will have to wrap their own server around the library.
 
+# Package Managers
+
+You can download and install LevelDB using the [vcpkg](https://github.com/Microsoft/vcpkg) dependency manager:
+
+    git clone https://github.com/Microsoft/vcpkg.git
+    cd vcpkg
+    ./bootstrap-vcpkg.sh
+    ./vcpkg integrate install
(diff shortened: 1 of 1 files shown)


pwnall:
Thank you for maintaining the vcpkg packaging and for your proposal!

In general, I don't want to end up maintaining per-packager configuration or docs. 10 lines of instructions is a bit too much.

I think it'd be reasonable to have an `Installation` section above `Building` that mentions package managers and installing from source as a backup strategy. Alternatively, a new first paragraph at the beginning of `Building` could state that this project was packaged by many package managers, such as ..., and the easiest way to use it is to install an existing package.

In either alternative of this counter-proposal, vcpkg would be listed as one of the package managers, with a link. The link would be your preferred landing page, which could contain installation instructions similar to what you're proposing here.

WDYT?

sipsorcery:
+1 for including the vcpkg instructions in the readme. As a Windows C++ developer vcpkg is the first approach I attempt for a new library. Fallback is library specific install/build instructions (in fairness to leveldb the documented cmake instructions are also very straight forward).

Issue #755, opened 2019-11-28
Unicode handling in new windows environment


laanwj:
The new windows environment uses the *A functions instead of the *W ones:
```c++
  size_t error_text_size = ::FormatMessageA(
    ScopedHandle handle = ::CreateFileA(
        ::CreateFileA(filename.c_str(), desired_access, share_mode,
        ::CreateFileMappingA(handle.get(),
    ScopedHandle handle = ::CreateFileA(
    ScopedHandle handle = ::CreateFileA(
    return GetFileAttributesA(filename.c_str()) != INVALID_FILE_ATTRIBUTES;
    HANDLE dir_handle = ::FindFirstFileA(find_pattern.c_str(), &find_data);
    } while (::FindNextFileA(dir_handle, &find_data));
    if (!::DeleteFileA(filename.c_str())) {
    if (!::CreateDirectoryA(dirname.c_str(), nullptr)) {
    if (!::RemoveDirectoryA(dirname.c_str())) {
    if (!::GetFileAttributesExA(filename.c_str(), GetFileExInfoStandard,
    if (::MoveFileA(from.c_str(), to.c_str())) {
    if (::ReplaceFileA(to.c_str(), from.c_str(), /*lpBackupFileName=*/nullptr,
    ScopedHandle handle = ::CreateFileA(
    if (!GetTempPathA(ARRAYSIZE(tmp_path), tmp_path)) {
```
I think these use what the process code page happens to be, which might not be UTF-8. I don't claim to understand Windows unicode handling, so please tell me I'm wrong, but in our [previous Windows environment](https://github.com/bitcoin-core/leveldb/blob/bitcoin-fork/util/env_win.cc) we used

```c++
void ToWidePath(const std::string& value, std::wstring& target) {
	wchar_t buffer[MAX_PATH];
	MultiByteToWideChar(CP_UTF8, 0, value.c_str(), -1, buffer, MAX_PATH);
	target = buffer;
}

void ToNarrowPath(const std::wstring& value, std::string& target) {
	char buffer[MAX_PATH];
	WideCharToMultiByte(CP_UTF8, 0, value.c_str(), -1, buffer, MAX_PATH, NULL, NULL);
	target = buffer;
}
```
along with the *W wide variants of Windows API functions. 

I'm seeing some downstream test failures regarding file names, and I suspect this might be the cause.

sipsorcery:
[...]

PR #760, opened 2019-12-11
Adds unicode support to Windows environment.

sipsorcery:
Currently the Windows environment uses the *A ANSI Win32 API calls for file system operations. This precludes the ability of consuming applications to pass paths with unicode characters. A detailed discussion is in #755. 

This PR swtiches the *A ANSI calls to the *W wide unicode string calls along with the conversion methods from standard UTF-8 strings to the UTF-16 multi-byte strings requires by the Win32 *W API functions.


util/env_windows_test.cc
```diff
@@ -1,65 +1,131 @@
-// Copyright (c) 2018 The LevelDB Authors. All rights reserved.
```
laanwj: I think something went wrong here, it's rewriting the entire file

sipsorcery: Thx, will check.

MarcoFalke:
Needs rebase?

jonsimantov:
Hi, is there any update to this PR? It would be very helpful if it were merged.

cucumber12121:
ping

PR #766, opened 2019-12-29
FileMetaData: fix init number

pavel-pimenov:


krishnacx (commented):
Better to include comments for members (refs and number).

PR #769, opened 2020-01-07
Minor proposal on line 58

kant:


krishnacx (commented):
Removing duplicate default would suffice.
"The default will build for x86. For 64-bit run:"

PR #771, opened 2020-01-10
Introduce EnvOptions

### LINKED ISSUE #768: Crashes on Snappy decompression (mmap/MapViewOfFile issues?)
I'm seeing crashes during Snappy decompression that I suspect are due to disk failures triggered by memory mapped file reads.

```
[ 00 ] snappy::SnappyDecompressor::DecompressAllTags<snappy::SnappyArrayWriter>
[ 01 ] snappy::RawUncompress
[ 02 ] leveldb::ReadBlock
[ 03 ] leveldb::Table::BlockReader
[ 04 ] leveldb::`anonymous namespace'::TwoLevelIterator::InitDataBlock
[ 05 ] leveldb::`anonymous namespace'::TwoLevelIterator::Seek
[ 06 ] leveldb::`anonymous namespace'::MergingIterator::Seek
[ 07 ] leveldb::`anonymous namespace'::DBIter::Seek
```

A similar issue has been reported a long time ago https://github.com/google/leveldb/issues/221, but it was closed without a definitive solution. The issue is not on `snappy` -- as mentioned in #221. That code is thoroughly exercised in Chromium.

But the File implementation used by `env_chromium.cc` doesn't rely on memory mapping -- https://cs.chromium.org/chromium/src/base/files/file_win.cc?sq=package:chromium&g=0&l=77 -- so I suspect these crashes are related to the native LevelDB `Env` implementations using `mmap` on POSIX and `MapViewOfFile` on Windows.

Is it possible to configure LevelDB to not use memory mapping? That would allow me to investigate further. Would a PR adding this be welcome? Should this be a compile-time flag or a runtime option passed to `Env::Default()`?

philix:
EnvOptions allows users to configure the `Env` without having to roll their own Windows/POSIX Env implementations.

`EnvPosix` supports the two options -- `readonly_mmap_files_limit` and `max_open_files`.

`EnvWindows` supports only `readonly_mmap_files_limit`.

The ability to disable memory mapping of files, allow LevelDB users to mitigate crashes due to unreliable storage media (fixes #768).


pwnall:
Sorry, I don't think EnvOptions belongs in the LevelDB API. I think it leads to a slippery slope where we're expected to get the default Env implementations to meet everyone's needs, and we spend a disproportionate amount of time working on them.

There's nothing wrong with forking and customizing our Env implementations.

1) `leveldb::Env`, the API between Env implementations and the rest of LevelDB, is stable.
2) You don't need to track our changes in your fork. If your fork works with your toolchain and OS, it's fine the way it is.

philix:
@pwnall thanks for taking a look! I had little hope of this being accepted upstream, but I tried anyways since I need to do this change on my side anyways.

I would need Windows and POSIX implementations of `Env`, so I prefer to have this commit in my fork and merge back google/leveldb changes to `env_windows.cc`/`env_posix.cc` and make the adaptations necessary to keep my changes working.

PR #774, opened 2020-01-10
Enable -Werror=conditional-uninitialized if available

Empact:
And fix two trivial local violations - initialized to the empty or
failing value.

Note, in the current code these are false positives as the called
functions do set the value in all cases, but these initializations
remove the reliance on that external property, and enable
automatic detections of future violations.

Motivated by a desire to enable this warning in a project
that builds leveldb as a dependency.


practicalswift:
Very good! Thanks for fixing this @Empact!

Would be really nice to get rid of these warnings when building Bitcoin Core!

PR #778, opened 2020-02-18
Update CONTRIBUTING.md

travinkeith:
Missing a key word in the second sentence. Without "through", there is no direction of where the "jump" is going. Though it's implied, it's better if it's in the sentence.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,7 +1,7 @@
 # Contributing
 
 We'd love to accept your code patches! However, before we can take them, we
-have to jump a couple of legal hurdles.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -1,7 +1,7 @@
 # Contributing
 
 We'd love to accept your code patches! However, before we can take them, we
-have to jump a couple of legal hurdles.
+have to jump through a couple of legal hurdles.
```
cmumford: Thx for the fix, but I think this is a mixed metaphor. One would either "jump through hoops" or "jump over hurdles".

PR #784, opened 2020-03-09
Delete useless code

caozhiyi:
the if is always true.


caozhiyi:
@googlebot I signed it!

PR #789, opened 2020-03-22
Make the judgment of whether the block needs to searched more accurate.

ShooterIT:
If there are only six characters, it will not be possible to have a complete record, so we don't need to search from the beginning of the block. But now we will not be searched only if it is less than six characters left.


ShooterIT:
@cmumford Please have a look if you have time

PR #797, opened 2020-04-12
Update CMakeLists.txt

caojunlin:
3.14 version is very available


pwnall:

We try to keep the *minimum* as low as possible. You should be able to build the project on newer versions of CMake as well.

PR #798, opened 2020-04-24
Remove unnecessary status check

lntotk:
In Table::Open function, we already judge status just after Decode footer, so the status judge before read index block is duplicate and unnessary,  we can remove it.


lntotk:
@googlebot I signed it!

lntotk:
is the mac clang configuration have some error?

pwnall:
And to answer the question above.. the CI was fixed recently. Rebasing on top of master would've given you a green check here.

At this point, however, it's better to leave this PR as-is. Messing with it may confuse the bot.

PR #800, opened 2020-04-28
fix build error

seanachao:
fix error when building , the error is " undefined reference to `leveldb::ConcurrentTest::K"


pwnall:

A similar fix was just published from our internal repository -- https://github.com/google/leveldb/commit/23d67e7c1f4396919bd0c73c0eced13a0dac37f3

Issue #770, opened 2020-01-09
Making error on Debian 10


akudryav:
During compilation I've received following error:
[ 60%] Linking CXX executable env_test
/usr/bin/ld: lib/libgmock.so: undefined reference to `testing::internal::ParseInt32(testing::Message const&, char const*, int*)'
collect2: error: ld returned 1 exit status

How can I fix it?
![Снимок экрана 2020-01-09 в 21 12 34](https://user-images.githubusercontent.com/2817119/72093449-6e90a080-3325-11ea-9e20-9d3f7cd505d7.png)

sfwtoms2030:
del  update and re-make

zhjwpku:
@akudryav pls try this patch.
https://github.com/google/leveldb/pull/806

k15tfu:
There is also one more error on macOS:
```
[ 65%] Linking CXX shared library libbenchmark_main.dylib
Undefined symbols for architecture x86_64:
  "benchmark::Initialize(int*, char**)", referenced from:
      _main in benchmark_main.cc.o
  "benchmark::RunSpecifiedBenchmarks()", referenced from:
      _main in benchmark_main.cc.o
  "benchmark::ReportUnrecognizedArguments(int, char**)", referenced from:
      _main in benchmark_main.cc.o
  "benchmark::internal::InitializeStreams()", referenced from:
      ___cxx_global_var_init in benchmark_main.cc.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

xloem:
I ran into this issue and #806 worked to resolve it for me.  `make test` passes fine.

PR #806, opened 2020-05-24
Fix the tests linking error with -DBUILD_SHARED_LIBS=ON

zhjwpku:
When build with BUILD_SHARED_LIBS=ON, the tests links with gmock
error:

  /usr/bin/ld: lib/libgmock.so: undefined reference to `testing::internal::ParseInt32(testing::Message const&, char const*, int*)'

which is reported by issue: https://github.com/google/leveldb/issues/770

This patch sets BUILD_SHARED_LIBS=OFF before enter googletest subdirecory,
and restores the previous value later. By doing this, gtest and gmock
will be built as static libraries.

I really think this is a bug in googletest CMakeLists, but I didn't dig
into it too much.

The solution might not be that elegant, but it works, maybe we can find
a better way later.

Signed-off-by: Zhao Junwang <zhjwpku@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -296,7 +296,10 @@ if(LEVELDB_BUILD_TESTS)
   set(build_gmock ON)
 
   # This project is tested using GoogleTest.
+  set(BUILD_SHARED_LIBS_SAVED "${BUILD_SHARED_LIBS}")
+  set(BUILD_SHARED_LIBS OFF)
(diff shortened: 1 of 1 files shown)


pwnall:

We're not currently interested in contributions to the build configuration. We now have a paragraph explaining the situation in  https://github.com/google/leveldb/blob/master/README.md#contributing-to-the-leveldb-project

PR #809, opened 2020-06-03
Fix finding Snappy

haraldF:
Just doing a check_library_exists is not good enough, we also need to
find the include path. Add a standard FindSnappy.cmake as suggested by
the CMake docs at https://cmake.org/cmake/help/latest/manual/cmake-developer.7.html#find-modules

This also takes standard CMake variables like CMAKE_PREFIX_PATH into
consideration and eases builds that have Snappy installed in a custom
directory. Also eases cross-compiling because it'll handle standard
paths better.

Closes google/leveldb#685


dktapps:
this breaks snappy compilation: https://travis-ci.org/github/google/leveldb/jobs/694435316#L3893

the `HAVE_SNAPPY` macro is needed in the source

haraldF:
> this breaks snappy compilation: https://travis-ci.org/github/google/leveldb/jobs/694435316#L3893
> 
> the `HAVE_SNAPPY` macro is needed in the source

Thanks for the hint, I indeed didn't check the `configure_file` below. Fixed by setting HAVE_SNAPPY internal CMake variable.

dktapps:
Something is still wrong: https://travis-ci.org/github/google/leveldb/jobs/696182885#L3877

haraldF:
> Something is still wrong: https://travis-ci.org/github/google/leveldb/jobs/696182885#L3877

Should be finally fixed now, I see the benchmark is using snappy and the generated `port_config.h` is also correct.

haraldF:
@dktapps hey, is there any way to merge this? All checks are green...

dktapps:
@haraldF I'm not a maintainer, I just happened to drive by because I was looking for a better Snappy CMake config for a fork I'm working on.

PR #820, opened 2020-07-12
add barrier-free variant of find_greater_or_equal

classhe:
add barrier-free variant of find_greater_or_equal

### DIFF
--- a/db/skiplist.h
+++ b/db/skiplist.h
@@ -117,7 +117,8 @@ class SkipList {
   //
   // If prev is non-null, fills prev[level] with pointer to previous
   // node at "level" for every level in [0..max_height_-1].
-  Node* FindGreaterOrEqual(const Key& key, Node** prev) const;
+  Node* FindGreaterOrEqual(const Key& key, Node** prev,
+                           bool barrier_free = false) const;
 
   // Return the latest node with a key < key.
   // Return head_ if there is no such node.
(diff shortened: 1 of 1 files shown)


classhe:
@googlebot I signed it!

pwnall:
WDYT about making barrier_free a template argument, to make sure it's always a constant? Otherwise, there's a (minor) risk that we end up with a conditional jump here, which would most likely erase any benefits obtained from removing the barrier.

TBH I'm not sure this will be a net-benefit. This will generate more code, adding presure to the L1 code cache. This change looks most attractive on ARM, where barriers are expensive... but ARMs tend to have smaller caches too. I'd appreciate it if you ran benchmarks so we'd know what kind of differences we're looking at. I expect the results will come out as a wash, in which case I'll defer to the `TODO`'s author and accept the change.

db/skiplist.h
```diff
@@ -259,12 +260,12 @@ bool SkipList<Key, Comparator>::KeyIsAfterNode(const Key& key, Node* n) const {
 
 template <typename Key, class Comparator>
 typename SkipList<Key, Comparator>::Node*
-SkipList<Key, Comparator>::FindGreaterOrEqual(const Key& key,
-                                              Node** prev) const {
+SkipList<Key, Comparator>::FindGreaterOrEqual(const Key& key, Node** prev,
```
pwnall, line 263: Please make this templated on barrier_free -- the template argument will end up being `bool BarrierFree = false`.

db/skiplist.h
```diff
@@ -338,7 +339,7 @@ void SkipList<Key, Comparator>::Insert(const Key& key) {
   // TODO(opt): We can use a barrier-free variant of FindGreaterOrEqual()
   // here since Insert() is externally synchronized.
   Node* prev[kMaxHeight];
-  Node* x = FindGreaterOrEqual(key, prev);
+  Node* x = FindGreaterOrEqual(key, prev, true);
```
pwnall, line 342: `true` -> `/* BarrierFree= */ true`

PR #825, opened 2020-07-27
fix typo in snapshot.h

sundb:
change "direcly" to "directly"


sundb:
@googlebot I signed it!

PR #831, opened 2020-08-07
add DB::Get definition to avoid link error when using gcc4.8

luohaha:
When I using gcc4.8 to build libleveldb.a，and link it to the example code ( https://github.com/google/leveldb/blob/master/doc/index.md )，the compiler complain ：

g++ -o db_writer db_writer.cpp /root/leveldb/libleveldb.a -std=c++11 -lpthread
/tmp/cc60A3F6.o: In function 'main':
db_writer.cpp:(.text+0x255): undefined reference to `leveldb::DB::Get(leveldb::ReadOptions const&, leveldb::Slice const&, std::string*)'
collect2: error: ld returned 1 exit status

But testing with LLVM version 8.1.0 and gcc7.4 the result is fine. I think it is better to be friendly to those weird compilers:)

### DIFF
--- a/db/db_impl.cc
+++ b/db/db_impl.cc
@@ -1473,6 +1473,11 @@ Status DB::Delete(const WriteOptions& opt, const Slice& key) {
   return Write(opt, &batch);
 }
 
+Status DB::Get(const ReadOptions& options, const Slice& key, 
+               std::string* value) {
+  return Get(options, key, value);
+}
+
 DB::~DB() = default;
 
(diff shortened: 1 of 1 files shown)


pwnall:
DB::Get() is a pure virtual, so the linker shouldn't expect an implementation.

It seems like you're trying to work around a bug in an old compiler. I'm afraid you'll have to carry this patch in your own fork, as we won't support this.

PR #850, opened 2020-11-27
Use 'N' mode in fopen() to disable handle inheritance on Windows

### LINKED ISSUE #623: Leaking file descriptors to child processes
If a process using leveldb fork/execs child processes, these child processes will  inherit leveldb's file handles, including `LOG`, `LOCK`, and `MANIFEST-*`, and most worryingly `*.log`.  This seems dangerous, as it could result in inadvertent database corruption.

For example, the contract between some of these children and scripts which they go on to spawn is for certain kinds of errors messages to be returned via a pipe on on file descriptor 6 - which is inherited to be the `LOG` file.  At the moment they are correctly closing the existing descriptor and reopening at the correct location, but it's easy to imagine a bug by which one of these child processes would start writing garbage (from leveldb's point of view) to the file descriptor.

The correct solution (for kernel 2.6.23 and higher) should be to add `O_CLOEXEC` to every `open` call, and for glibc 2.7 and higher add `e` to the mode string for `fopen`.  For older kernel/glibc versions (which, arguably, there shouldn't be much need to support any more), one can work around the lack of functionality, setting `fcntl(fd, F_SETFD, FD_CLOEXEC)` will work, although it can still leak in multithreaded programs.  See for example how [jemalloc](https://github.com/jemalloc/jemalloc/blob/5e23f96dd4e4ff2847a85d44a01b66e4ed2da21f/src/prof.c#L1678) handles this case.

Others have noted this, for example (including a patch which it looks like they're applying locally): https://jira.apache.org/jira/browse/MESOS-8913

I can't imagine a safe, legitimate use case for children to inherit these handles, but if someone else can imagine such a situation then perhaps it could be made a database creation option.

k15tfu:
Fixes #623


k15tfu:
@cmumford Hi! Could you take a look?

cmumford:
Hi @k15tfu 

Thx for the PR (with tests). Even though the docs state that only one process can access a leveldb, I'm fine with adding a protection to help protect against accidental multi-process use. One problem is that this relies on a POSIX extension that is only available on Windows - please correct me if I'm wrong. I'm worried that this would wind up with application bugs that only show up on non-Windows platforms.

k15tfu:
@cmumford Hi! This is not for protection. Without this change, the log file handle leaks to child processes and therefore I cannot close DB and move/delete it until all handles are closed. So in my case it fails until the child process exits.

In this PR I change only Windows specific files and use platform-dependent `N` mode for `fopen()` (https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen?view=msvc-140). A similar change was already done for `fopen()` in `env_posix.cc` (https://github.com/google/leveldb/pull/624/files#diff-6e3e34528a780e660b519271c86d81c55519dcaf2fb7965fd0b232ae5cdcf0c5L677-R688), so now there should be no difference.

util/env_windows_test.cc
```diff
@@ -2,12 +2,67 @@
...
+
 #include "gtest/gtest.h"
 #include "leveldb/env.h"
 #include "port/port.h"
 #include "util/env_windows_test_helper.h"
 #include "util/testutil.h"
 
+namespace {
+
+void GetOpenHandles(std::unordered_set<HANDLE>* open_handles) {
```
pwnall, line 17: (GetOpenFileDescriptors in env_posix_test.cc is a bad example here, sorry. Any chance you'd be interested in fixing that test in a separate PR? If not, it's OK, I can follow up.)

Prefer return values to out params. https://google.github.io/styleguide/cppguide.html#Inputs_and_Outputs

In this case, the function should return a std::unordered_set.

util/env_windows_test.cc
```diff
@@ -29,7 +84,7 @@ TEST_F(EnvWindowsTest, TestOpenOnRead) {
   ASSERT_LEVELDB_OK(env_->GetTestDirectory(&test_dir));
   std::string test_file = test_dir + "/open_on_read.txt";
 
-  FILE* f = std::fopen(test_file.c_str(), "w");
+  FILE* f = std::fopen(test_file.c_str(), "wN");
```
pwnall, line 87: N is a non-standard extension that most developers will not be familiar with. Can you add a comment explaining it? Here is some draft text.

N is a Windows-specific extension that makes the file handle non-inheritable by child processes. Without this option, open handles in child processes may prevent the database from getting deleted.

pwnall (changes requested):
Thank you for the explanation, and for pointing out that we're already using the non-standard "e" fopen() mode in env_posix.cc. Based on this, it seems reasonable to make the same change for Windows.

The code looks good overall. Happy to LGTM after the changes below are made.

util/env_windows_test.cc
k15tfu: I can't make them return non-void because I used ASSERT_* macros.

util/env_windows_test.cc
k15tfu: Added in WindowsEnv::NewLogger. Should I add this here too?

k15tfu:
@pwnall Friendly ping.

k15tfu:
@pwnall Friendly ping.

k15tfu:
@pwnall Friendly ping.

k15tfu:
@pwnall @cmumford Friendly ping.

k15tfu:
@pwnall @cmumford Friendly ping.

ForNeVeR:
Yup that's correct. Too sad that the corresponding test changes are still in this (stale) PR :(

PR #854, opened 2020-11-30
Fixed fprintf of 64-bit value.

cmumford:
A recent change (ed78107) switched from `%8d` to `%8zu` for fprintf'ing a uint64_t value which does not work on all platforms. This should fix platforms by using the PRIu64 constant.

### DIFF
--- a/db/db_test.cc
+++ b/db/db_test.cc
@@ -5,6 +5,7 @@
 #include "leveldb/db.h"
 
 #include <atomic>
+#include <cinttypes>
 #include <string>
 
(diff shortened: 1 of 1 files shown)


naisteph24 (commented):
bc1q4at02q3f6dvakpzfsmjr3wjsgx87eng2yq6ll2

naisteph24 (approved):
bitcoin:1GFegds3uDk2mLUrt9SYaFwVFxtoTrVJk2

PR #859, opened 2020-12-11
Introduce BuildDBFromSstFiles()

songzhi:
This feature is inspired by RocksDB, see in:
https://rocksdb.org/blog/2017/02/17/bulkoad-ingest-sst-file.html

This patch implements a simplified version which learns from RocksDB. We
can create sst files, then build the DB from them. The simplification
introduces some restrictions:
  1. db directory must be empty;
  2. all sst files keys must be not overlapping.


pwnall:
Thank you for this contribution!

Unfortunately, I don't see the value of maintaining this feature in LevelDB. Bulk loading where the keys are already sorted (so the disjoint SST constraint can be met) is a fairly rare case.

johnedmonds:
Google Kythe accomplishes this by [writing the table files](https://github.com/kythe/kythe/blob/5599bd84deb62f61d58c83dcb37079023fce82d7/kythe/go/serving/pipeline/beamio/leveldb.go#L316), then [writing the manifest](https://github.com/kythe/kythe/blob/5599bd84deb62f61d58c83dcb37079023fce82d7/kythe/go/serving/pipeline/beamio/leveldb.go#L143). If the .ldb files are disjoint, LevelDB leaves them alone and everything works magically. If not, LevelDB starts sorting them. It looks like they pull this off with some 3rd-party LevelDB implementation written in Go.

I'd just like to give my +1 that this is a feature I'd find useful.

Issue #868, opened 2021-01-11
Breaking build - Optimize leveldb block seeks to utilize the current iterator location


swiftomkar:
Hi, 
I was trying to build LevelDB with the updates to master and noticed that my build is breaking.
On a quick look, I noticed that there was a commit made a couple of hours ago and that might be the reason for the breaking build. 
Here is the link to that [commit](https://github.com/google/leveldb/commit/8cce47e450b365347769959c53b8836ef0216df9)
Here's the failure log from my machine: 
```
/home/odesai/leveldb/benchmarks/db_bench.cc:152:40: error: use of deleted function ‘std::atomic<long unsigned int>::atomic(const std::atomic<long unsigned int>&)’
  152 |   mutable std::atomic<size_t> count_ = 0;
      |                                        ^
In file included from /home/odesai/leveldb/benchmarks/db_bench.cc:7:
/usr/include/c++/9/atomic:824:7: note: declared here
  824 |       atomic(const atomic&) = delete;
      |       ^~~~~~
/usr/include/c++/9/atomic:828:17: note:   after user-defined conversion: ‘constexpr std::atomic<long unsigned int>::atomic(std::atomic<long unsigned int>::__integral_type)’
  828 |       constexpr atomic(__integral_type __i) noexcept : __base_type(__i) { }
      |                 ^~~~~~
make[2]: *** [CMakeFiles/db_bench.dir/build.make:89: CMakeFiles/db_bench.dir/benchmarks/db_bench.cc.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:245: CMakeFiles/db_bench.dir/all] Error 2
make: *** [Makefile:141: all] Error 2
```

I also noticed that travis has the same build failure: https://travis-ci.org/github/google/leveldb/jobs/753958518

Is there anything that I am missing?

swiftomkar:
@pwnall

zhxq:
I can confirm this on a macOS 11.1 computer, with slightly different notices (but of the same error), showing 
```
/Users/zhxq/Downloads/testing/leveldb/benchmarks/db_bench.cc:133:7: error: 'Compare' overrides a member function but is not marked 'override' [-Werror,-Winconsistent-missing-override]
  int Compare(const Slice& a, const Slice& b) const {
      ^
/Users/zhxq/Downloads/testing/leveldb/include/leveldb/comparator.h:28:15: note: overridden virtual function is here
  virtual int Compare(const Slice& a, const Slice& b) const = 0;
[...]

PR #869, opened 2021-01-11
fix issue #868

swiftomkar:
This PR intends to fix issue #868 
Seems like it was a deleted copy constructor that was getting called. 
May be an issue with come compilers. 
Here is some more detail from the build toolchain:
```
-- The C compiler identification is GNU 9.3.0
-- The CXX compiler identification is GNU 9.3.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
...
```


swiftomkar:
@googlebot I signed it!

PR #875, opened 2021-02-18
hotfix#fix the range predication in log_reader.cc

guoxiangCN:
When a block cell cannot fit a content of 3 bytes in length, the write will be filled with zeros. I think there is a lack of judgment on whether offset is within the filling interval.


guoxiangCN:
> @googlebot I signed it!

PR #881, opened 2021-03-04
Don't include C++ headers in extern C

paulb777:
This fixes a build failure in an upcoming version of the Apple toolchain.

More details at https://github.com/firebase/firebase-ios-sdk/pull/7563

### DIFF
--- a/include/leveldb/c.h
+++ b/include/leveldb/c.h
@@ -40,15 +40,15 @@
 #ifndef STORAGE_LEVELDB_INCLUDE_C_H_
 #define STORAGE_LEVELDB_INCLUDE_C_H_
 
-#ifdef __cplusplus
-extern "C" {
-#endif
-
 #include <stdarg.h>
 #include <stddef.h>
 #include <stdint.h>
 
 #include "leveldb/export.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
 
 /* Exported types */
 


pwnall (approved):
LGTM, thank you for the fix! This PR will get merged once it's applied to the internal repository.

(me trying to rationalize the change) From an ODR perspective, this makes sense.

extern "C" may impact code generation, so it's not OK to #include the same header with and without  an extern "C" scope. We're #including the headers here without extern "C" in other C++ files, so we can't use extern "C" here.

paulb777:
Thanks for the quick turnaround!

BTW, here's the full error log from Xcode before the fix:

```
While building module 'leveldb' imported from /Users/paulbeusterien/Library/Developer/Xcode/DerivedData/AsyncAwaitFirebase-auvgxgocidaxqpeyrdxbpsbgaski/SourcePackages/checkouts/firebase-ios-sdk/Firestore/core/src/local/leveldb_util.h:24:
In file included from <module-includes>:1:
/Users/paulbeusterien/Library/Developer/Xcode/DerivedData/AsyncAwaitFirebase-auvgxgocidaxqpeyrdxbpsbgaski/SourcePackages/checkouts/leveldb/include/leveldb/c.h:51:1: error: import of C++ module 'leveldb' appears within extern "C" language linkage specification [-Wmodule-import-in-extern-c]
#include "leveldb/export.h"
^
/Users/paulbeusterien/Library/Developer/Xcode/DerivedData/AsyncAwaitFirebase-auvgxgocidaxqpeyrdxbpsbgaski/SourcePackages/checkouts/leveldb/include/leveldb/c.h:44:1: note: extern "C" language linkage specification begins here
extern "C" {
^
1 error generated.
In file included from /Users/paulbeusterien/Library/Developer/Xcode/DerivedData/AsyncAwaitFirebase-auvgxgocidaxqpeyrdxbpsbgaski/SourcePackages/checkouts/firebase-ios-sdk/Firestore/core/src/local/leveldb_util.cc:17:
/Users/paulbeusterien/Library/Developer/Xcode/DerivedData/AsyncAwaitFirebase-auvgxgocidaxqpeyrdxbpsbgaski/SourcePackages/checkouts/firebase-ios-sdk/Firestore/core/src/local/leveldb_util.h:24:10: fatal error: could not build module 'leveldb'
#include "leveldb/slice.h"
 ~~~~~~~~^~~~~~~~~~~~~~~~~
2 errors generated
```

pwnall:
@paulb777 Thank you very much for listing the error message here, for posterity!

1) I'm not following Apple toolchains as closely as you are. If you happen to notice when Travis CI gets support for this new toolchain, giving us a heads-up / contributing a config change would mean that we won't accidentally break you in the future :smile: 

2) The Xcode output mentions `C++ module 'leveldb'`. I wonder if this references [C++20 modules](https://en.cppreference.com/w/cpp/language/modules) or some Xcode construct called "modules".

Just to be clear, we haven't thought of C++20 modules for leveldb. I'd appreciate it if you discuss with us before exposing the leveldb API in a C++20 module, so we can reason through any tradeoffs together. Thank you!

paulb777:
1. We've mostly migrated from Travis to GitHub Actions for CI. Either way, we discovered this investigating a future Swift feature async/await that is only available now in an unreleased future toolchain. Setup details in the Firebase PR mentioned above.

2. These aren't C++ modules - they're Objective C modules necessary for consuming libraries from Swift and optionally from Objective C. Even though Firebase is only accessing leveldb from Objective C and C++, I don't think there is a way to build without modules if the app is Swift.

Wangtt185 (commented):
好

PR #883, opened 2021-03-16
Remove dead code

storagezhang:
1. Remove dead code.

   In `table/table_test.cc`,  `TestRandomAccess`'s conditions always be false because **static const bool kVerbose** has not been changed since initialization.

   ```c++
   static const bool kVerbose = false;
   ...;
   if (kVerbose) std::fprintf(stderr, "---\n");
   ...;
   if (kVerbose) std::fprintf(stderr, "Next\n");
   ...;
   if (kVerbose) std::fprintf(stderr, "SeekToFirst\n");
   ...;
   if (kVerbose)
               std::fprintf(stderr, "Seek '%s'\n", EscapeString(key).c_str());
   ...;
   if (kVerbose) std::fprintf(stderr, "Prev\n");
   ...;
   if (kVerbose) std::fprintf(stderr, "SeekToLast\n");
   ```

2. Fix some typos in comments.

### DIFF
--- a/db/snapshot.h
+++ b/db/snapshot.h
@@ -25,7 +25,7 @@ class SnapshotImpl : public Snapshot {
   friend class SnapshotList;
 
   // SnapshotImpl is kept in a doubly-linked circular list. The SnapshotList
-  // implementation operates on the next/previous fields direcly.
+  // implementation operates on the next/previous fields directly.
   SnapshotImpl* prev_;
   SnapshotImpl* next_;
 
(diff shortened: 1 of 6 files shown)


ghemawat:
Thanks for the suggestion. However, I would prefer to keep the kVerbose related code around. The reason we have the "kVerbose" declaration there is so that it is easy to turn on extra logging when needed to diagnose a test failure. I would like to preserve that ability.

PR #851, opened 2020-11-27
Fixup for compiling benchmark

ehds:
DB benchmark targets can not find `gtest` and `gmock` when set  `LEVELDB_BUILD_TESTS` to `OFF` but `LEVELDB_BUILD_BENCHMARKS` to `ON`.

PR #884, opened 2021-03-17
Fix benchmark build error when test is disabled.

ls4154:
Building the leveldb benchmark requires both google test and google benchmark, which are included only when `LEVELDB_BUILD_TESTS` is enabled.


ls4154:
@googlebot I signed it!

classhe:
same with #851

to4ka56 (approved):

PR #893, opened 2021-04-05
Delete an unnecessary forward declaration

myccccccc:
already include the header file `arena.h`, delete forward declaration for `class Arena`
https://github.com/google/leveldb/blob/f57513a1d6c99636fc5b710150d0b93713af4e43/db/skiplist.h#L34-L39

### DIFF
--- a/db/skiplist.h
+++ b/db/skiplist.h
@@ -36,8 +36,6 @@
 
 namespace leveldb {
 
-class Arena;
-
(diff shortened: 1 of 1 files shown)


pwnall (approved):
LGTM

Arena is defined in the `util/arena.h` header, which is included above. The definition is used by template code in this header.

myccccccc:
.

Issue #898, opened 2021-04-26
Invalid key may not be dropped during DoCompactionWork


coyorkdow:
In the `Status DBImpl::DoCompactionWork(CompactionState* compact)` we have codes below to clean the redundant keys:
```c++
      if (!has_current_user_key ||
          user_comparator()->Compare(ikey.user_key, Slice(current_user_key)) !=
              0) {
        // First occurrence of this user key
        current_user_key.assign(ikey.user_key.data(), ikey.user_key.size());
        has_current_user_key = true;
        last_sequence_for_key = kMaxSequenceNumber;
      }

      if (last_sequence_for_key <= compact->smallest_snapshot) {
        // Hidden by an newer entry for same user key
        drop = true;  // (A)
      } else if (ikey.type == kTypeDeletion &&
                 ikey.sequence <= compact->smallest_snapshot &&
                 compact->compaction->IsBaseLevelForKey(ikey.user_key)) {
        // For this user key:
        // (1) there is no data in higher levels
        // (2) data in lower levels will have larger sequence numbers
        // (3) data in layers that are being compacted here and have
        //     smaller sequence numbers will be dropped in the next
        //     few iterations of this loop (by rule (A) above).
        // Therefore this deletion marker is obsolete and can be dropped.
        drop = true;
      }
```
From my understanding of the above codes, if there are multi user keys appears, with sequence numbers are not greater than smallest snapshot. Then only the user key with sequence number most closed to smallest snapshot will be kept, while others will be dropped (Similar logic for deletion type, too). It works fine definitely. But consider such scenario:
```
Put('key', 'v1'); \\ seq 0
Put('key', 'v2'); \\ seq 1
GetSnapshot(); \\ snapshot 0
Put('key', 'v3'); \\ seq 2
Put('key', 'v4'); \\ seq 3
GetSnapshot(); \\ snapshot 1, assume it is the latest snapshot
......
```
What we truly need is the <'key', 'v2'> (for snapshot 0) , <'key', 'v4'> (for snapshot 1), and the latest Put/Delete of 'key'. However, after running DoCompactionWork, only <'key', 'v1'> be dropped, while the other invalid kv pairs (seq 2, and so on) are still stored in LSM-tree, wasting disk space.
I wonder if there exists an better garbage collection algorithm to avoid such problem.

coyorkdow:
I create a pull request to slove it. https://github.com/google/leveldb/pull/900

philix:
[...]

PR #900, opened 2021-04-27
Drop more invalid entries for same key.

coyorkdow:
In the old approach to dropping, there may be invalid entries that can't be removed, see more in https://github.com/google/leveldb/issues/898
For each snapshot, there may exists multiple entries for the same key which sequence number less or equal to it, what we really need is the closest one. A vector of snapshots (MaxSequenceNumber will be appended to the end) is used to help determine if an entry should be dropped. When a user key occurs first time, snapshots reverse iterator points to rbegin. If we found current key's sequence number <= iterator, then it's the closest one to previous snapshot.
I add a new test `HiddenValuesAreRemoved2` (rename: `HiddenValuesAreRemoved` to `HiddenValuesAreRemoved1`) to test such case.

### DIFF
--- a/db/db_impl.cc
+++ b/db/db_impl.cc
@@ -63,18 +63,11 @@ struct DBImpl::CompactionState {
 
   explicit CompactionState(Compaction* c)
       : compaction(c),
-        smallest_snapshot(0),
         outfile(nullptr),
         builder(nullptr),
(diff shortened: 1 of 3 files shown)


coyorkdow:
@pwnall @cmumford Could you please look at it?

PR #902, opened 2021-05-08
Make table cache non-copyable

ehds:
Make table cache non-copyable

### DIFF
--- a/db/table_cache.h
+++ b/db/table_cache.h
@@ -22,6 +22,10 @@ class Env;
 class TableCache {
  public:
   TableCache(const std::string& dbname, const Options& options, int entries);
+
+  TableCache(const TableCache&) = delete;
+  TableCache& operator=(const TableCache&) = delete;
+
   ~TableCache();
 
   // Return an iterator for the specified file number (the corresponding


pwnall (approved):
LGTM

This brings the code in compliance with the C++ style guide, which says that classes should explicitly state if they're copyable/movable.

TableCache is already non-copyable/non-movable because it defines an explicit constructor (so the compiler will not auto-generate a copy constructor) and has const members (so the compiler will not auto-generate a copy assignment). So, expecting no build breakages.

ehds:
> LGTM
> 
> This brings the code in compliance with the C++ style guide, which says that classes should explicitly state if they're copyable/movable.
> 
> TableCache is already non-copyable/non-movable because it defines an explicit constructor (so the compiler will not auto-generate a copy constructor) and has const members (so the compiler will not auto-generate a copy assignment). So, expecting no build breakages.

Explicit constructor will not prevent compiler generating a default copy constructor (refer to [copy_constructor](https://en.cppreference.com/w/cpp/language/copy_constructor)). So explicitly deleted state is necessary.

Issue #904, opened 2021-05-15
wrong manifest_file_number_ in VersionSet::Recover


myccccccc:
in VersionSet::Recover, we set manifest_file_number_ to be the next_file_number_ that we recovered from manifest file
https://github.com/google/leveldb/blob/f57513a1d6c99636fc5b710150d0b93713af4e43/db/version_set.cc#L974
and we use this manifest_file_number_ to create a new manifest file when recover
https://github.com/google/leveldb/blob/f57513a1d6c99636fc5b710150d0b93713af4e43/db/version_set.cc#L808
this assumes that the next_file_number_ that we recovered from manifest file, is the most update to date next_file_number_ in VersionSet when we are about to create the new manifest file. 
this is dangerous, since there might be file not recorded in the old manifest file is using the next_file_number_ that we recovered from the old manifest file

PR #905, opened 2021-05-15
fix manifest_file_number_ when recover

myccccccc:
#904 
in VersionSet::Recover, we set manifest_file_number_ to be the next_file_number_ that we recovered from manifest file
https://github.com/google/leveldb/blob/f57513a1d6c99636fc5b710150d0b93713af4e43/db/version_set.cc#L974
and we use this manifest_file_number_ to create a new manifest file when recover
https://github.com/google/leveldb/blob/f57513a1d6c99636fc5b710150d0b93713af4e43/db/version_set.cc#L808
this assumes that the next_file_number_ that we recovered from manifest file, is the most update to date next_file_number_ in VersionSet when we are about to create the new manifest file. 
this is dangerous, since there might be file not recorded in the old manifest file is using the next_file_number_ that we recovered from the old manifest file

### DIFF
--- a/db/recovery_test.cc
+++ b/db/recovery_test.cc
@@ -112,6 +112,8 @@ class RecoveryTest : public testing::Test {
 
   uint64_t FirstLogFile() { return GetFiles(kLogFile)[0]; }
 
+  uint64_t ManifestFile() { return GetFiles(kDescriptorFile)[0]; }
+
   std::vector<uint64_t> GetFiles(FileType t) {
     std::vector<std::string> filenames;
     EXPECT_LEVELDB_OK(env_->GetChildren(dbname_, &filenames));
@@ -320,6 +322,31 @@ TEST_F(RecoveryTest, MultipleLogFiles) {
   ASSERT_EQ("bar2", Get("foo"));
   ASSERT_EQ("world", Get("hello"));
   ASSERT_EQ("there", Get("hi"));
+}
+
+TEST_F(RecoveryTest, MultipleLogFilesNotReuseManifest) {
+  ASSERT_LEVELDB_OK(Put("foo", "bar"));
+  Close();
+  ASSERT_EQ(1, NumLogs());
+
+  // Make a bunch of uncompacted log files.
+  uint64_t old_log = FirstLogFile();
+  MakeLogFile(old_log + 1, 1000, "hello", "world");
+  MakeLogFile(old_log + 2, 1001, "hi", "there");
+  MakeLogFile(old_log + 3, 1002, "foo", "bar2");
+
+  // Recover create new manifest and check that all log files were processed.
+  // and check that the new manifest file number should be larger than log files.
+  Options opts;
+  opts.reuse_logs = false;
+  Open(&opts);
+  ASSERT_EQ("bar2", Get("foo"));
+  ASSERT_EQ("world", Get("hello"));
+  ASSERT_EQ("there", Get("hi"));
+  ASSERT_GT(ManifestFile(), old_log);
+  ASSERT_GT(ManifestFile(), old_log + 1);
+  ASSERT_GT(ManifestFile(), old_log + 2);
+  ASSERT_GT(ManifestFile(), old_log + 3);
 }
 
 TEST_F(RecoveryTest, ManifestMissing) {
--- a/db/version_set.cc
+++ b/db/version_set.cc
@@ -805,6 +805,7 @@ Status VersionSet::LogAndApply(VersionEdit* edit, port::Mutex* mu) {
     // No reason to unlock *mu here since we only hit this path in the
     // first call to LogAndApply (when opening the database).
     assert(descriptor_file_ == nullptr);
+    manifest_file_number_ = NewFileNumber();
     new_manifest_file = DescriptorFileName(dbname_, manifest_file_number_);
     edit->SetNextFile(next_file_number_);
     s = env_->NewWritableFile(new_manifest_file, &descriptor_file_);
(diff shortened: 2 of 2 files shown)


myccccccc:
if we reuse manifest file, manifest_file_number_ will be set correctly by parsing manifest filename in VersionSet::ReuseManifest
https://github.com/google/leveldb/blob/f57513a1d6c99636fc5b710150d0b93713af4e43/db/version_set.cc#L1022
else manifest_file_number_ will be given a NewFileNumber() when creating this new manifest file

myccccccc:
@pwnall PTAL

myccccccc:
@cmumford @pwnall PTAL

pwnall:
Could you add a test that covers the failure mode this PR is claiming to fix? The test should fail without the code changes here, and pass after the code changes are applied.

ghemawat:
Can you explain the problem you foresee in more detail?

My thinking is that the current code works fine and prevents a problem where in some situations we may exhaust all available file numbers.

(1) Why I think the current code is fine:  a new manifest file is not considered committed until its name has been written to CURRENT. Therefore it should be safe to ignore and possibly overwrite files that have a number larger than the one recorded in CURRENT.

(2) Why the proposed change may be unsafe: Consider a directory that contains a stray file with a large number L that got created by mistake. Before this change: we would ignore L and create a new manifest file at some number < L.  After a long period of operation, leveldb's next_file_number_ may reach L and at that point will overwrite L.

After this change: we will start creating manifest files at L+1. And if L happens to be close to UINT64_MAX, we will soon run out of available file numbers.

myccccccc:
> Can you explain the problem you foresee in more detail?
> 
> My thinking is that the current code works fine and prevents a problem where in some situations we may exhaust all available file numbers.
> 
> (1) Why I think the current code is fine: a new manifest file is not considered committed until its name has been written to CURRENT. Therefore it should be safe to ignore and possibly overwrite files that have a number larger than the one recorded in CURRENT.
> 
> (2) Why the proposed change may be unsafe: Consider a directory that contains a stray file with a large number L that got created by mistake. Before this change: we would ignore L and create a new manifest file at some number < L. After a long period of operation, leveldb's next_file_number_ may reach L and at that point will overwrite L.
> 
> After this change: we will start creating manifest files at L+1. And if L happens to be close to UINT64_MAX, we will soon run out of available file numbers.

(1) there might be some log files that are not recorded in the manifest file(DB is down before writing a versionedit into manifest) before we recover. While we are recovering, we first read the manifest file and get the next_file_number_, but the logs' file number should be equal or larger than this next_file_number_ since they are not recoreded in the manifest. If we use this next_file_number_ as the new manifest file number, it will be equal to one log file number. Shouldn't all files have different file numbers?

(2) If we really have a stray file with a large number L that got created by mistake, then this DB can't work. Because when we are recovering we call MarkFileNumberUsed
https://github.com/google/leveldb/blob/f57513a1d6c99636fc5b710150d0b93713af4e43/db/version_set.cc#L1026-L1030
after that next_file_number_ will be L+1

(3) The changed code is more readable

ghemawat:
> (1) there might be some log files that are not recorded in the manifest file(DB is down before writing a versionedit into manifest) before we recover. While we are recovering, we first read the manifest file and get the next_file_number_, but the logs' file number should be equal or larger than this next_file_number_ since they are not recoreded in the manifest. If we use this next_file_number_ as the new manifest file number, it will be equal to one log file number. Shouldn't all files have different file numbers?

The code is intentionally saving and recording the last valid log file number in CURRENT. Files with any other numbers are assumed to be invalid (they could have partial data for example).

> (2) If we really have a stray file(kLogFile) with a large number L that got created by mistake, then this DB can't work. Because when we are recovering we call MarkFileNumberUsed when we meet a log file not recorded in the manifest

MarkFileNumberUsed arguments come from edits read from a log file. The problem I was worried about was somebody doing
    touch "log_file_name for UINT64_MAX".

As pwnall@ asked before, perhaps you have run into a problem and give details from that, or show a test case that demonstrates the problem you are worried abouyt.

myccccccc:
> > (1) there might be some log files that are not recorded in the manifest file(DB is down before writing a versionedit into manifest) before we recover. While we are recovering, we first read the manifest file and get the next_file_number_, but the logs' file number should be equal or larger than this next_file_number_ since they are not recoreded in the manifest. If we use this next_file_number_ as the new manifest file number, it will be equal to one log file number. Shouldn't all files have different file numbers?
> 
> The code is intentionally saving and recording the last valid log file number in CURRENT. Files with any other numbers are assumed to be invalid (they could have partial data for example).
> 
> > (2) If we really have a stray file(kLogFile) with a large number L that got created by mistake, then this DB can't work. Because when we are recovering we call MarkFileNumberUsed when we meet a log file not recorded in the manifest
> 
> MarkFileNumberUsed arguments come from edits read from a log file. The problem I was worried about was somebody doing
> touch "log_file_name for UINT64_MAX".
> 
> As pwnall@ asked before, perhaps you have run into a problem and give details from that, or show a test case that demonstrates the problem you are worried abouyt.

first make a bunch of uncompacted log files as the existing test does
https://github.com/google/leveldb/blob/f57513a1d6c99636fc5b710150d0b93713af4e43/db/recovery_test.cc#L279-L299
then perform manifest number check

myccccccc:
@pwnall @ghemawat  PTAL

myccccccc:
@pwnall @ghemawat  PTAL

ghemawat:
> first make a bunch of uncompacted log files as the existing test RecoveryTest.MultipleLogFiles does

That's an internal test that is testing a lower-level implementation detail. And the proposed fix suffers from the problems I already mentioned twice.

I am going to close this pull request since I have seen no evidence of a problem without this change and have at least a potential scenario of of a problem that can occur after the change.

PR #917, opened 2021-06-25
MANIFEST file size will be huge after long time running

suuyaoo:
MANIFEST file size will be huge after long time running, rotate MANIFEST when MANIFEST file write size exceed write buffer size to avoid this


Krishna7557:
Hi ,

I hope issue 899 and issue 299 are getting addressed in this commit. We are having few queries about the fix:

As per the implementation, once the Manifest file reaches write buffer size older file will be closed and descriptor_log_ = nullptr;
      descriptor_file_ = nullptr;
are set to null which triggers the below code frequently:

if (descriptor_log_ == nullptr) {
    // No reason to unlock *mu here since we only hit this path in the
    // first call to LogAndApply (when opening the database).
    assert(descriptor_file_ == nullptr);
    new_manifest_file = DescriptorFileName(dbname_, manifest_file_number_);
    edit->SetNextFile(next_file_number_);
    s = env_->NewWritableFile(new_manifest_file, &descriptor_file_);
    if (s.ok()) {
      descriptor_log_ = new log::Writer(descriptor_file_);
      s = WriteSnapshot(descriptor_log_);
    }
  }
Does writing snapshot frequently have any performance impacts?
what is the content of descriptor_log_ ? Is it only the manifest file content or entire database content?

Regards,
krishna

Krishna7557:
@suuyaoo can you comment on the above queries?

Regards,
Krishna

suuyaoo:
@Krishna7557  Each snapshot write is triggered after write_buffer_size has been accumulated, even if the actual manifest file size has exceeded write_buffer_size, which ensures that snapshot writes are not triggered too often, and the overall performance impact is minimal.
The content of descriptor_log_ is only the manifest file content.

db/version_set.cc
```diff
@@ -797,6 +798,20 @@ Status VersionSet::LogAndApply(VersionEdit* edit, port::Mutex* mu) {
...
 
+  if (descriptor_log_ != nullptr) {
+    if (manifest_file_size_hint_ >= options_->write_buffer_size) {
+      Log(options_->info_log, "MANIFEST size %ld exceed, switch to next file\n",
+          long(manifest_file_size_hint_));
+      delete descriptor_log_;
+      delete descriptor_file_;
+      descriptor_log_ = nullptr;
+      descriptor_file_ = nullptr;
+      manifest_file_number_ = manifest_file_number_ + 1;
```
hellowangsai, line 809: Call NewFileNumber() here?

hellowangsai: You should set manifest_file_number_ after SetCurrentFile() successes. Otherwise, the DeleteObsoleteFiles() may delete the current valid manifest file.

suuyaoo: This can not happen because DeleteObsoleteFiles and LogAndApply both run in the background compact thread.

mvtm-dn:
Any changes?

superboyiii:
Any changes?

PR #919, opened 2021-07-01
Fixed incorrect limit in env_posix_test.cc

wineway:
```
  // Exhaust the RandomAccessFile mmap limit. This way, the test
  // RandomAccessFile instance below is backed by a file descriptor, not by an
  // mmap region.
  leveldb::RandomAccessFile* mmapped_files[kReadOnlyFileLimit] = {nullptr};
  for (int i = 0; i < kReadOnlyFileLimit; i++) {
    ASSERT_LEVELDB_OK(env_->NewRandomAccessFile(file_path, &mmapped_files[i]));
  }
```
In above code; we should use kMMapLimit to exhaust the RandomAccessFile mmap;
Otherwise when kMMapLimit is not equals to kReadOnlyFileLimit, this test would failed

### DIFF
--- a/util/env_posix_test.cc
+++ b/util/env_posix_test.cc
@@ -243,8 +243,8 @@ TEST_F(EnvPosixTest, TestCloseOnExecRandomAccessFile) {
   // Exhaust the RandomAccessFile mmap limit. This way, the test
   // RandomAccessFile instance below is backed by a file descriptor, not by an
   // mmap region.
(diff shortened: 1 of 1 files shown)


wineway:
@googlebot I signed it!

pwnall (approved):
LGTM.

This change aligns the code with the comment above it. Thank you for the fix!

PR #922, opened 2021-07-25
Update .gitignore

SmartKeyerror:
Add CLion auto-generated directories to .gitignore, may be more friendly to CLion users.


pwnall:

While this makes sense in general, I'll draw the line to tools used by the LevelDB developers here at Google.

PR #928, opened 2021-08-08
Fix comments position for Version::Get()

ehds:
Fix comments position


pwnall (approved):
LGTM. Thank you for the fix!

PR #938, opened 2021-09-28
Reduce code redundancy

theidexisted:
SingletonEnv shares almost the same code with NoDestructor, except for the test-related flag.


pwnall:
Agreed with the finding. However, I'd like to reduce the dependencies of our Env implementations on other internal headers. Ideally, users could copy-paste our Env implementations and evolve them, without any non-public dependencies on LevelDB.

PR #941, opened 2021-10-09
Windows: Prevent LOG handle from being inherited by subprocesses

dktapps:
I recently encountered a problem with this because Windows doesn't allow
files to be deleted when there's open handles to them.

Other files opened by leveldb are not affected because by and large they
are using CreateFileA, which does not allow inheritance when
lpSecurityAttributes is null (ref:
https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilea)

However, fopen() _does_ allow inheritance, and it needs to be expressly
disabled.
https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen?view=msvc-160

### DIFF
--- a/util/env_windows.cc
+++ b/util/env_windows.cc
@@ -622,7 +622,7 @@ class WindowsEnv : public Env {
   }
 
   Status NewLogger(const std::string& filename, Logger** result) override {
-    std::FILE* fp = std::fopen(filename.c_str(), "w");
+    std::FILE* fp = std::fopen(filename.c_str(), "wN");
     if (fp == nullptr) {
       *result = nullptr;
       return WindowsError(filename, ::GetLastError());


dktapps:
@googlebot I signed it!

pwnall (approved):
LGTM. Thank you for your contribution!

"N" is not standardized (POSIX, C2011, etc.) but it is documented in MSDN: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen?view=msvc-170 -- I think this is sufficient for Windows-specific code.

This PR will get imported in our internal repository, and then a bot will merge it here.

dktapps:
Awesome, thanks.

PR #944, opened 2021-10-19
Tool for restore data from damaged bases

proller:
When you have only *.ldb files without indexes this can help.
Also can try read damaged *.ldb without footer.
Restore data from lost dir.

Or your app have exceptions like
/build/leveldb-s1RAd7/leveldb-1.22/table/table_builder.cc:99: void leveldb::TableBuilder::Add(const leveldb::Slice&, const leveldb::Slice&): Assertion `r->options.comparator->Compare(key, Slice(r->last_key)) > 0' failed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -286,6 +286,12 @@ add_executable(leveldbutil
 )
 target_link_libraries(leveldbutil leveldb)
 
+add_executable(leveldbrestore
+  "db/leveldbrestore.cc"
+)
+target_link_libraries(leveldbrestore leveldb)
+
+
 if(LEVELDB_BUILD_TESTS)
   enable_testing()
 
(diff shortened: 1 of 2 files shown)


pwnall:
Thank you for this contribution! It is clear to me that you've spent a lot of time putting this together, and I appreciate that.

Unfortunately, this is not in a good state to be merged. Tactically, the code doesn't follow the Google C++ style guide, and doesn't have testing coverage. At a higher level, I think that recovery code should be landed as improvements to leveldb::RepairDB(), not as a separate tool. The PR description lists a few improvements that could be proposed as separate PRs.

This is a lot of extra work. In most cases, it probably makes most sense for you to keep using your tool to solve the problems you've run into.

proller:
Style guide was got from tools around.
I solved all my problems with corrupted bases with this tool.
This is only one tool in whole internet for restoring data, by closing this pr you hide last hope from other peoples.

Issue #863, opened 2020-12-18
Make background thread stoppable


nshishov:
Hi,

Current version of LevelDB has detached background thread that can't be stopped and each environment calls `std::abort()` in destructor.
This behavior prevents using of LevelDB in plugin dlls that can be started and stopped when main application still runs.

I think it is better to allow background thread to finish tasks in queue and then exit, when environment is being destroyed. That will make database loadable and unloadable.

pwnall:
Thank you for the clear explanation!

We may be willing to consider a PR that fixes this problem, depending on the complexity it introduces in the codebase.

If we end up not accepting the fix, you can still use it in your own codebase. LevelDB is designed to work with custom Env implementations. The ones supplied in the project are intended to be good enough for most cases, and can be used as starting points when you need to write your own.

dlannan-fmad:
Hi. Just wondering if this might make it work? We are using C api. 
Use ```leveldb_create_default_env();``` to create a default env. Then remove the environment with ```leveldb_env_destroy(leveldb_env_t*)`` which would correctly destroy the Env object? 
And then this could be repeated?
The problem I see that could occur, is the BackgroundMainThread may close during operations? Like compaction? 
Im still learning my way thought leveldb, so let me know if I have this all wrong.

PR #948, opened 2021-11-03
Stoppable environment

nshishov:
Implementation of #863 
If Environment is set in options it will be stopped correctly

```
std::unique_ptr<leveldb::Env> levelDBEnvironment{ leveldb::Env::DefaultNotStatic() };
leveldb::Options options {};
options.env = levelDBEnvironment.get();
```

### DIFF
--- a/include/leveldb/env.h
+++ b/include/leveldb/env.h
@@ -64,6 +64,14 @@ class LEVELDB_EXPORT Env {
   // The result of Default() belongs to leveldb and must never be deleted.
   static Env* Default();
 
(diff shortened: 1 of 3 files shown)


nshishov:
@googlebot I signed it!

pwnall:

We're not currently interested in supporting this lifetime requirement in the environments provided by this repository -- there's quite a bit of extra complexity that we'd have to figure out how to cover with tests, and most applications can embed LevelDB without this requirement.

I think this is a very reasonable approach in a custom environment.

PR #951, opened 2021-11-14
Remove <pthread.h> include

philix:
Since https://github.com/google/leveldb/commit/d177a026, pthreads is not used directly and leveldb relies on the C++11 threading primitives.

### DIFF
--- a/util/env_posix.cc
+++ b/util/env_posix.cc
@@ -4,7 +4,6 @@
 
 #include <dirent.h>
 #include <fcntl.h>
-#include <pthread.h>
 #include <sys/mman.h>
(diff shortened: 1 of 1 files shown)


pwnall:
Thank you for looking into this! I think that the #include removal is valid, but the `find_package` is still needed.

philix:
> Thank you for looking into this! I think that the #include removal is valid, but the find_package is still needed.

True. Although pthread is not used in the source code it's still needs to be linked. I updated the diff.

pwnall (approved):
LGTM

Thank you for your contribution! This PR will go through our internal repository, and eventually get merged by a bot.

PR #952, opened 2021-11-15
Fixtestexplain

baiqiubai:
the first minor compact level0 has no files,the will push level is kMaxMemCompactLevel


baiqiubai:
@googlebot I signed it!

PR #957, opened 2021-12-02
Use FindFile to implement Compaction::IsBaseLevelForKey

baiqiubai:
we can add flag to identify we should use which comparator in FindFile. use user_comparator to implement 
Compaction::IsBaseLevelForKey. Do this we can  use binary search.  we don need level_ptrs_. I test


baiqiubai:
@pwnall can you take a look please?

PR #960, opened 2021-12-05
rm redundant code: SetNextFile has already been called before in this…

ericuni:
`edit->SetNextFile` has already been called before, there is no need to call it again.


pwnall (approved):
LGTM

VersionEdit::SetNextFile() was already called on line 789 above, which is guaranteed to execute in all the cases where line 809 executes.

The `VersionEdit::next_file_number_` data member of the `edit` local is not changed between 789 and 809. The local is only used

* on line 790, by VersionEdit::SetLastSequence(), which changes a different data member.
* on line 795, by a Builder::Apply(), which doesn't mutate its VersionEdit argument, and doesn't store a reference to any of its data.

PR #962, opened 2021-12-09
optimize Compaction::ShouldStopBefore

rainliang000:
Mostly, the overlap grandparents files total size is small than 10*max_file_size.
So I think we can check the overlaps grandparents file total size first. 
If match it, quickly return false to continue.


rainliang000:
CLA signed.

rainliang000:
could you please help remove the pr, thanks.

Issue #946, opened 2021-10-25
Limiter class bug in env_posix


kiner-shah:
https://github.com/google/leveldb/blob/c5d5174a66f02e66d8e30c21ff4761214d8e4d6d/util/env_posix.cc#L96

If `Release` is called multiple times (instead of calling once), the count can increase `max_acquires`. This can later lead to violation of the constraint to limit the resources to `max_acquires`.

Suggested fix:
1. Store `max_acquires` passed in the constructor in a `const` variable, say `resource_limit_`.
2. Check inside `Release()` if `acquires_allowed_` is greater than `resource_limit_`. If yes, ignore adding.

duststarr:
It is only used in the same source file.

kiner-shah:
@duststarr it doesn't matter, it's a bug which needs to be fixed.

ghemawat:
This is not a bug. It is a responsibility of the user of Limiter to match
one Release per Acquire. Allowing redundant releases would be unsafe.

Consider the following sequence of operations:

Thread 1: Acquire
Thread 2: Acquire
Thread 1: Release
Thread 1: Release (redundant)
Thread 2: Release

The redundant release by Thread 1 (if allowed and supported, instead of
being considered a bug) would prematurely cancel Thread 2's Acquire.

kiner-shah:
@ghemawat I think this is a bug due to the design of the Limiter class. The issue is "who owns what" and requirement is to allow only the owners to release the resources only as many times they have acquired. So in your example, if Thread 1 has already released the resource, next time calling Release again should have no effect (somehow mark that Thread 1 has released the resource and check next time if it's already marked, and if yes, ignore the Release).

Regarding your statement of allowing users to ensure they are not calling Release more than once, it should ideally be a two-way responsibility: user should avoid calling Release twice, but by chance (or mistake) if Release is called twice, Limiter should automatically make sure that it doesn't affect the availability of resources.

@duststarr appreciate your efforts for creating a pull request :-) but I think my suggested solution isn't a proper one. It probably will be more complex than I initially thought.

duststarr:
I agree that it's not necessay to make this class too complex.

pwnall:
335876a1335c765f818ae10d9c4d18f563cdfce5 adds some asserts that clarify the responsibilities of the `Limiter` users.

PR #963, opened 2021-12-15
[fix]Limiter class bug in env_posix #946

duststarr:
As the issue #946 .
Add a check for max_acquires when Release() called to ignore redundant calls.

### DIFF
--- a/util/env_posix.cc
+++ b/util/env_posix.cc
@@ -73,7 +73,8 @@ Status PosixError(const std::string& context, int error_number) {
 class Limiter {
  public:
   // Limit maximum number of resources to |max_acquires|.
-  Limiter(int max_acquires) : acquires_allowed_(max_acquires) {}
+  Limiter(int max_acquires) 
+      : acquires_allowed_(max_acquires), max_acquires_(max_acquires) {}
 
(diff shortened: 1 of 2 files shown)


ghemawat:
There is no way to safely allow redundant releases, since the new release
may be non-redundant and the redundant pair of acquires may have happened
earlier. E.g.,

thread 1:
   acquire
   release
   release (redundant)

thread 2:
   acquire
   release

The redundant release in thread 1 already broke things for thread 2 by
cancelling its acquire early.

As you mentioned in the previous discussion, this is a local class, and
there is a clear responsibility for the owner to match up acquire and
releases. So let's stick with the old code.

dencesun:
good

pwnall:
We could do something like https://github.com/pwnall/leveldb/commit/0bc209d4639f7b9b06514b6b53b6551268bcbbc6 to make it obvious that calling Limiter::Release() too many times is a bug in the Limiter client.

pwnall:
Landed the above in 335876a1335c765f818ae10d9c4d18f563cdfce5

duststarr:
Good solution.

PR #965, opened 2021-12-31
`std::memory_order::memory_order_relaxed` -> `std::memory_order_relaxed`

ShawnZhong:
`std::memory_order::memory_order_relaxed` is removed since C++ 20: https://en.cppreference.com/w/cpp/atomic/memory_order. 

Running `cmake -DCMAKE_CXX_STANDARD=20 .. && cmake --build .` gives the following error:

```
leveldb/util/env_posix.cc:890:35: error: no member named 'memory_order_relaxed' in 'std::memory_order'; did you mean 'std::memory_order_relaxed'?
    assert(!env_initialized_.load(std::memory_order::memory_order_relaxed));
                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                  std::memory_order_relaxed
```

This PR fixed this issue.

### DIFF
--- a/util/env_posix.cc
+++ b/util/env_posix.cc
@@ -870,7 +870,7 @@ class SingletonEnv {
  public:
   SingletonEnv() {
 #if !defined(NDEBUG)
-    env_initialized_.store(true, std::memory_order::memory_order_relaxed);
+    env_initialized_.store(true, std::memory_order_relaxed);
 #endif  // !defined(NDEBUG)
     static_assert(sizeof(env_storage_) >= sizeof(EnvType),
                   "env_storage_ will not fit the Env");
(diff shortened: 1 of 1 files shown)


carlossn11 (approved):

pwnall (approved):
Thank you for this contribution!

This PR will be merged by a bot after it makes it way through our internal repositories.

I applied the same changes to env_windows.cc and verified that tests still pass in https://github.com/pwnall/leveldb/runs/4749567817. I changed the commit description to a summary of the PR description.

Please don't change your branch -- the bot will merge the commit approved here.

mikehardy:
@pwnall and maybe @a-sully - many apologies for the direct tag, but this was never released. It is currently causing build failures for us over in react-native-firebase (which wraps Google Firebase) because in a react-native context they have started defining NDEBUG during the build, which opens these non-C++-confirmant lines to the compiler and they fail to compile

Is there any way we could get a leveldb 1.24 release with this fix in it 🙏 ?

PR #973, opened 2022-01-12
optimize includes

julianreisch:
We applied our tool for optimizing includes to the leveldb project. It ensures for each file that header files are included if and only if they are used. As can be seen, many includes can be removed completely or moved from .h to .cc files so that in total, there are many more deletions than additions. Also, this reduces the number of lines of code after the preprocessor (in .o files) by 15.3% which results in a shorter compilation time. We computed only 114.4 sec user time on average on our machine, compared to 134.5 sec without the optimization - a reduction by 14.9%.

We remark that we applied the clang format only to the parts where we modified include directives and nowhere else. This explains why sometimes there are empty lines between the include directives now.

### DIFF
--- a/util/testutil.cc
+++ b/util/testutil.cc
@@ -2,9 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file. See the AUTHORS file for names of contributors.
 
-#include "util/testutil.h"
-
-#include <string>
+#include "leveldb/slice.h"
 
 #include "util/random.h"
 
--- a/benchmarks/db_bench.cc
+++ b/benchmarks/db_bench.cc
@@ -2,10 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file. See the AUTHORS file for names of contributors.
 
-#include <sys/types.h>
-
 #include <atomic>
-#include <cstdio>
 #include <cstdlib>
 
 #include "leveldb/cache.h"
@@ -13,8 +10,10 @@
 #include "leveldb/db.h"
 #include "leveldb/env.h"
 #include "leveldb/filter_policy.h"
+#include "leveldb/iterator.h"
+#include "leveldb/options.h"
 #include "leveldb/write_batch.h"
-#include "port/port.h"
+
 #include "util/crc32c.h"
 #include "util/histogram.h"
 #include "util/mutexlock.h"
--- a/db/autocompact_test.cc
+++ b/db/autocompact_test.cc
@@ -2,10 +2,11 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file. See the AUTHORS file for names of contributors.
 
-#include "gtest/gtest.h"
 #include "db/db_impl.h"
+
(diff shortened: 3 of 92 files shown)


julianreisch:
My company Synoptics (https://groups.google.com/g/synoptics) signed the CLA. Can please a repo manager rescan?

util/testutil.cc
```diff
@@ -2,9 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file. See the AUTHORS file for names of contributors.
 
-#include "util/testutil.h"
```
pwnall, line 5: This is incorrect.

Every .cc file starts out by including its corresponding header. This lets the compiler that

1. Definitions in the .cc match the declarations in the .h file
2. The .h file compiles without any external #includes.

util/testutil.cc
```diff
@@ -2,9 +2,7 @@
...
-
-#include <string>
```
pwnall, line 7: `std::string` is used below. Per IWYU, `<string>` must be included.

pwnall (commented):
The tool used here doesn't seem to understand the Google C++ style guide and IWYU (Include What You Use).

pwnall:
Thank you for your contribution! Unfortunately, we cannot accept this as-is.

There's definitely some value here -- some of the proposed additions and deletions are valid, and the compilation speed-ups seem like a net win. However, the cost of manually checking every proposed addition and deletion is too high for us right now.

julianreisch:
@pwnall Thank you for the feedback. We enabled our tool to apply the IWYU rules so that 1.) headers a.h are remain included in their a.cc files and 2.) each file includes its uses. You can find the new result here: 

https://github.com/google/leveldb/compare/master...julianreisch:optimize-includes-applying-the-iwyu-rule?expand=1

The diff is now significantly smaller (19 additions and 88 deletions) and the number of preprocessed lines still decreases by 10%.
Would you reconsider the contribution?

pwnall:
@julianreisch I skimmed through a few files and the suggestions seem like a good step forward. Thank you!

FWIW, the include list is not complete -- for example, `util/status.cc` uses `std::memcpy` and therefore should include `<cstring>`. But I find that all the deleted #includes are indeed unused, and all the additions are correct. Can you please update this PR?

julianreisch:
@pwnall I have open the new PR #975 for the contribution. I hope that is what I was supposed to do.

julianreisch:
@pwnall With regard to your comment: We have spotted several includes that ought to be present according to IWYU but in order to keep the diff small, we decided not to include them directly if they were included transitively already. 

If you want to see the result if all uses yield their own direct include, please see: https://github.com/google/leveldb/compare/main...julianreisch:apply-iwyu-completely?expand=1
If we allow transitive includes at least for external files, see: https://github.com/google/leveldb/compare/main...julianreisch:apply-iwyu-partially?expand=1
(For example, in `util/stats.cc` the includes `cstdio`, `cstring`, `string` and `cassert` suffice even tough there are also uses to `features.h`, `new` and `memory`, but these files are included transitively).

PR #974, opened 2022-01-14
fix a typo in impl.md

fishmandev:
Fix a typo


cyber9unk (changes requested):
it was correct, should be "are".

PR #975, opened 2022-01-18
Optimizeincludes

julianreisch:
We applied our tool for optimizing includes to the leveldb project. According to the Google C++ style guide it ensures for each file that header files are included if and only if they are used. In order to keep the changes as small as possible, we do not replace transitive includes by direct ones. This optimization reduces the number of lines of code after the preprocessor (in .o files) by 10% which results in a shorter compilation time.

### DIFF
--- a/benchmarks/db_bench.cc
+++ b/benchmarks/db_bench.cc
@@ -3,6 +3,7 @@
 // found in the LICENSE file. See the AUTHORS file for names of contributors.
 
 #include <atomic>
+#include <cstdio>
 #include <cstdlib>
 
 #include "leveldb/cache.h"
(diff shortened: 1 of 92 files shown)


julianreisch:
@pwnall Hello, you suggested to update our original PR #973 which, however, is closed already. So I opened this one. I am just wondering if this is fine now.

Concerning the missing CLA: My company Synoptics (https://groups.google.com/g/synoptics) has signed it for me. Can you please rescan?

Thanks!

julianreisch:
@googlebot can you please rescan my CLA? My company Synoptics GmbH has signed it.

PR #980, opened 2022-01-21
Use macro instead of literal to declare namespace

iamazy:


iamazy:
@pwnall can you take a look please?

PR #988, opened 2022-02-18
Fix Windows Env file path handling

felipecrv:
After this fix, LevelDB will be able to open databases that are found
in a directory that contains special characters.

This happens a lot in practice if the user of a product using LevelDB
has special characters in their Windows account name. The user account
name is used to name that user's folder in the system where
application-specific data is stored.


cyber9unk (approved):
Checked, looks great.

PR #1000, opened 2022-03-27
Report corruption for invalid InternalKeys read from disk.

### LINKED ISSUE #966: ExtractUserKey() should be able to signal corruption
ExtractUserKey() has an assertion that its input size is at least 8 bytes, but the callers don't have checks corresponding to this assertion. This is fine under normal operation, but can lead to crashes when operating on corrupted files.

#103 documents such a crash. https://crbug.com/964851 explains how to create a corrupted file that will result in a crash.

We should audit all calls to ExtractUserKey() to see if the pre-condition is guaranteed to be met in non-obvious ways. If that's the case, we should document the reasoning. If not (my assumption), we should have ExtractUserKey() return a MaybeSlice (regular Slice plus invalid state), and bubble up ExtractUserKey() failures as corruption.

chjj:
Fixes #966.

Warning that I am not a C++ programmer and don't know how it works.

An example of my ignorance: does adding a "virtual" method to a class result in an ABI change? I imagine that a compiler might implement that as a function pointer on a struct or something.

I figure there's probably a better way to do the `IsInternal` check but I don't feel like researching C++ features at the moment.

### DIFF
--- a/db/version_edit.cc
+++ b/db/version_edit.cc
@@ -87,6 +87,7 @@ void VersionEdit::EncodeTo(std::string* dst) const {
 static bool GetInternalKey(Slice* input, InternalKey* dst) {
   Slice str;
   if (GetLengthPrefixedSlice(input, &str)) {
+    if (str.size() < 8) return false;
     return dst->DecodeFrom(str);
   } else {
     return false;
--- a/table/block.cc
+++ b/table/block.cc
@@ -162,6 +162,11 @@ class Block::Iter : public Iterator {
   }
 
   void Seek(const Slice& target) override {
+    if (comparator_->IsInternal() && target.size() < 8) {
+      CorruptionError();
+      return;
+    }
+
     // Binary search in restart array to find the last restart point
     // with a key < target
     uint32_t left = 0;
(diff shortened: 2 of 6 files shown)


db/version_edit.cc
```diff
@@ -87,6 +87,7 @@ void VersionEdit::EncodeTo(std::string* dst) const {
 static bool GetInternalKey(Slice* input, InternalKey* dst) {
   Slice str;
   if (GetLengthPrefixedSlice(input, &str)) {
+    if (str.size() < 8) return false;
     return dst->DecodeFrom(str);
```
chjj, line 91: Could also change InternalKey::DecodeFrom to check for 8 bytes I suppose.

table/block.cc
```diff
@@ -162,6 +162,11 @@ class Block::Iter : public Iterator {
   }
 
   void Seek(const Slice& target) override {
+    if (comparator_->IsInternal() && target.size() < 8) {
```
chjj, line 165: I don't think this check is actually necessary, but I put it here anyway.

chjj:
Been thinking about this more: if there is an ABI change, we can avoid it by converting the Comparator passed in Options to a child class (KeyComparator?) with internal linkage which has an `IsInternal()` method. InternalKeyComparator would then be made to inherit from KeyComparator.

Obviously, this would require some more changes (`sed 's/\<Comparator\>/KeyComparator/g'` in a lot of places, I'm sure). I'm willing to do that if the current PR is not acceptable.

We could also simply pass an option to Block::Iter signaling whether we're dealing with InternalKeys or user keys, but there's so many levels of iterators that option would need to pass through it's probably not worth the effort.

PR #1018, opened 2022-05-26
Update NEWS

Ajaygithubid:


cmumford:
Appears to be spam - just adding person's name.

PR #1032, opened 2022-06-18
Check for integer overflows involving ReadBlock.

chjj:


chjj:
Tacked one more commit on here for yet another overflow.

PR #1033, opened 2022-06-18
Fix undefined behavior in Block::Iter and Hash.

chjj:
This is part 3 of my work to harden leveldb against corruption and maliciously crafted databases (see also: #1032 and #1000).

`GetRestartPoint` currently blindly trusts the restart offsets at the end of the block, does some pointer arithmetic and then checks `p >= limit` after the fact.

If a restart offset exceeds the buffer size, the mere creation of `p` is undefined behavior and the comparison of `p >= limit` becomes meaningless. This could really cause issues on 32-bit (imagine a restart offset of `2^30` overflowing the pointer).

`Hash` also potentially creates an OOB pointer with `while (data + 4 <= limit)`.

### DIFF
--- a/table/block.cc
+++ b/table/block.cc
@@ -99,7 +99,8 @@ class Block::Iter : public Iterator {
 
   uint32_t GetRestartPoint(uint32_t index) {
     assert(index < num_restarts_);
-    return DecodeFixed32(data_ + restarts_ + index * sizeof(uint32_t));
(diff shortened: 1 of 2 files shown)


chjj:
Another instance of UB: FilterBlockReader was assigning `base_lg_` (used for a right shift) to `contents[n - 1]` without any validation. The most recent commit simply changes the assignment to `contents[n - 1] mod 64`.

PR #1034, opened 2022-06-19
Evict table file before archival in RepairTable.

chjj:
The `RepairTable` code currently renames the existing table file while the table cache still has an open file descriptor for it. This can cause sharing violations on windows (unless `FILE_SHARE_DELETE` is enabled on the handle).

In any case, it's probably better to avoid renaming open files in general, and in this case the open file isn't ever used again after `ArchiveFile` is called.

### DIFF
--- a/db/repair.cc
+++ b/db/repair.cc
@@ -313,6 +313,8 @@ class Repairer {
     }
     delete iter;
 
+    table_cache_->Evict(t.meta.number);
+
     ArchiveFile(src);
     if (counter == 0) {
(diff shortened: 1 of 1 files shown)


chjj:
Just an aside related to renaming files: we may also consider changing [WindowsEnv::RenameFile][1] to a simple call to `MoveFileExA(from, to, MOVEFILE_REPLACE_EXISTING)`.

[1]: https://github.com/google/leveldb/blob/e4ccaa0/util/env_windows.cc#L558

PR #1045, opened 2022-07-18
Add note for inactive maintenance

ayuishii:
There should be a note to communicate its current development state to developers.
This adds a note on its maintenance level and the changes we will accept.


pwnall (approved):
LGTM w/ comments addressed.

You may have an easier time submitting a CL against the internal repository and publishing it, instead of going through the PR merging process.

ayuishii:
Ah good point, thanks for the quick review. Will create new change in internal repository with comments addressed.

PR #1059, opened 2022-10-07
add bazel support

fansehep:
Signed-off-by: fan <yfan3763@gmail.com>

I think bazel is better than cmake anyway. cmake is too cumbersome.
Because nowadays i am learning leveldb. Sometimes, i want to add some tests in it. but cmake grammar is complicated. so i spend some time to add bazel build.


fansehep:
@Tesbarmix1  hi, thanks for your approved. and i have test ```bazel build``` in docker-ubuntu, docker-mint, archlinux. All of them have passed.
but i have not test it in Mac os. because i have no mac.

PR #1065, opened 2022-10-31
Update snapshot.h

ibrahimmemonn:


AnuguRakeshReddy (commented):
come on man. you better add 100 blank new lines to be a contributer

PR #1086, opened 2023-01-20
fix undefined behavior due to reading iterator data after modification in builder.cc

reznikovl:
Currently, the BuildTable() function tracks the largest key as the file is iterated over and adds it to the largest field of the FileMetadata at the end. However, the Slice returned from iter->key() refers to the same memory as the iterator, which is not guaranteed to be safe after the iterator is changed (ie, when iter->next() is called). As per the documentation in iterator.h:

`// Return the key for the current entry.  The underlying storage for
  // the returned slice is valid only until the next modification of
  // the iterator.
  // REQUIRES: Valid()
  virtual Slice key() const = 0;`

I believe this can cause undefined behavior, potentially adding junk data into meta->largest. To fix the memory bug and avoid the efficiency issue brought up in #819, I suggest seeking to the last position and then decoding that into the file, avoiding the issue.


felipecrv:
How can the `builder->Add` calls modify what `iter` sees?

db/builder.cc
```diff
@@ -30,14 +30,12 @@ Status BuildTable(const std::string& dbname, Env* env, const Options& options,
...
-      key = iter->key();
-      builder->Add(key, iter->value());
-    }
-    if (!key.empty()) {
-      meta->largest.DecodeFrom(key);
+      builder->Add(iter->key(), iter->value());
     }
+    
+    iter->SeekToLast();
+    meta->largest.DecodeFrom(iter->key());
```
felipecrv, line 38: Must check if `iter->Valid()` between the `SeekToLast` and `iter->key()` call.

db/builder.cc
```diff
@@ -30,14 +30,12 @@ Status BuildTable(const std::string& dbname, Env* env, const Options& options,
 
     TableBuilder* builder = new TableBuilder(options, file);
     meta->smallest.DecodeFrom(iter->key());
-    Slice key;
     for (; iter->Valid(); iter->Next()) {
-      key = iter->key();
-      builder->Add(key, iter->value());
-    }
-    if (!key.empty()) {
```
felipecrv, line 38: Unrelated to your change: this check assumes that `iter->Valid()` returned `false` due to having not more data in the iterator, but what if `iters->status()` is not OK? That's an error.

reznikovl:
Thank you for the response @felipecrv !

> How can the `builder->Add` calls modify what `iter` sees?
`builder->Add` won't modify it, but `iter->next()` might. Setting `key=iter.key()` shallow copies the `char*`, so if the underlying memory gets changed (such as when `iter.next, `key` will not be valid outside of the loop.

db/builder.cc
reznikovl: > Must check if `iter->Valid()` between the `SeekToLast` and `iter->key()` call.

Good point - will fix it!

PR #1089, opened 2023-01-23
Modify spelling errors: iff -> if

yangyang233333:


felipecrv:
`iff` is mathematical-speak for `if and only if`.

https://en.wikipedia.org/wiki/If_and_only_if

yangyang233333:
谢谢你的回答

PR #1104, opened 2023-03-28
Fix tests when run against ChromiumEnv

reillyeon:
There are a couple differences between ChromiumEnv and PosixEnv/WindowsEnv which cause test failures that are fixed (or at least patched over) in this change:

* NewSequentialFile() and NewRandomAccessFile() return Status::IOError rather than Status::NotFound when a file is not found, due to https://crbug.com/760362. This means a few tests need to expect a different error result.
* GetChildren() never returns the '.' or '..' entries.

### DIFF
--- a/db/db_test.cc
+++ b/db/db_test.cc
@@ -1722,8 +1722,13 @@ TEST_F(DBTest, DestroyEmptyDir) {
   ASSERT_TRUE(env.FileExists(dbname));
   std::vector<std::string> children;
   ASSERT_LEVELDB_OK(env.GetChildren(dbname, &children));
+#if defined(LEVELDB_PLATFORM_CHROMIUM)
+  // Chromium's file system abstraction always filters out '.' and '..'.
+  ASSERT_EQ(0, children.size());
+#else
   // The stock Env's do not filter out '.' and '..' special files.
   ASSERT_EQ(2, children.size());
+#endif  // defined(LEVELDB_PLATFORM_CHROMIUM)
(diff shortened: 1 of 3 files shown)


db/db_test.cc
```diff
@@ -1722,8 +1722,13 @@ TEST_F(DBTest, DestroyEmptyDir) {
   ASSERT_TRUE(env.FileExists(dbname));
   std::vector<std::string> children;
   ASSERT_LEVELDB_OK(env.GetChildren(dbname, &children));
+#if defined(LEVELDB_PLATFORM_CHROMIUM)
+  // Chromium's file system abstraction always filters out '.' and '..'.
```
pwnall, line 1726: Can you add a TODO pointing to the bug for aligning ChromiumEnv with the API?

util/env_test.cc
```diff
@@ -96,40 +96,45 @@ TEST_F(EnvTest, RunMany) {
   struct RunState {
```
pwnall, line 96: This is a very nice test improvement. Thank you!

I think this is independent of the Chromium-specific workarounds. No action needed for now, but if we need to revert / reland this CL for any reason, I'll probably ask you to pull these bits in a separate CL.

pwnall (changes requested):
I'll LGTM and push this through the internal repository as soon as you address the TODO comment. Thank you!

pwnall (approved):
LGTM, thank you for the revisions!

PR #1111, opened 2023-04-18
Check slice length in Footer::DecodeFrom()

reillyeon:
Without this check decoding the footer in Table::Open() can read uninitialized bytes from a buffer allocated on the stack if the file was unexpectedly short.

In practice this is probably fine since this function validates a magic number but MSan complains about branching on uninitialized data.


ailadas (commented):
table/format.cc

PR #1128, opened 2023-06-21
fix memory leak in kyoto cabinet benchmarks

WangTingZheng:
we should delete db_ in ~Benchmark() function


devarshishimpi (approved):

WangTingZheng:
@devarshishimpi Can you help me merge this pr to leveldb?

PR #1136, opened 2023-07-10
fix(hash): fix nullptr ub in hash function

WangTingZheng:
pass 0 int const char* will be nullptr. nullptr + 4 is ub,


chjj:
Already fixed in #1033.

PR #1028, opened 2022-06-09
fix macro HAVE_O_CLOEXEC when O_CLOEXEC not found

emptyx0:
port_config.h define HAVE_O_CLOEXEC is 0 when O_CLOEXEC not found in fcntl.h, so condition '#if defined(HAVE_O_CLOEXEC)' hit. It could be verified on centos5

PR #1149, opened 2023-10-01
Fix build break on system without O_CLOEXEC

stepinto:
On system without O_CLOEXEC, HAVE_O_CLOEXEC is defined as 0 in include/port/port_config.h, not undefined.  Therefore, the right way to test it is "#if HAVE_O_CLOEXEC" rather than "#if defined(...)".


hebasto:
Also see https://github.com/google/leveldb/pull/1028.

PR #1202, opened 2024-08-08
[jumbo] Add begin()/end() to Slice.

pkasting:
This allows this type to meet the requirements of e.g. std::ranges::range, which is necessary for it to work with the std::span range constructor, or the "non-legacy" constructor for Chromium's base::span.

Bug: none


huajie3030:
> This allows this type to meet the requirements of e.g. std::ranges::range, which is necessary for it to work with the std::span range constructor, or the "non-legacy" constructor for Chromium's base::span.
> 
> Bug: none

pkasting:
@a-sully Do you have permissions to review and land this?

Issue #1247, opened 2025-01-27
regression: does not compile because of C++17 construct


bertsky:
The following build problem was just introduced [here](https://github.com/google/leveldb/commit/302786e211d1f2e6fd260261f642d03a91e5922c):

```
In file included from leveldb/util/comparator.cc:14:
leveldb/./util/no_destructor.h: In constructor ‘leveldb::NoDestructor<InstanceType>::NoDestructor(ConstructorArgTypes&& ...)’:
leveldb/./util/no_destructor.h:24:24: error: ‘is_standard_layout_v’ is not a member of ‘std’; did you mean ‘is_standard_layout’?
   24 |     static_assert(std::is_standard_layout_v<NoDestructor<InstanceType>>);
      |                        ^~~~~~~~~~~~~~~~~~~~
      |                        is_standard_layout
leveldb/./util/no_destructor.h:24:70: error: expected primary-expression before ‘>’ token
   24 |     static_assert(std::is_standard_layout_v<NoDestructor<InstanceType>>);
      |                                                                      ^~
leveldb/./util/no_destructor.h:24:72: error: expected primary-expression before ‘)’ token
   24 |     static_assert(std::is_standard_layout_v<NoDestructor<InstanceType>>);
      |                                                                        ^
make[2]: *** [CMakeFiles/leveldb.dir/build.make:468: CMakeFiles/leveldb.dir/util/comparator.cc.o] Fehler 1
make[1]: *** [CMakeFiles/Makefile2:183: CMakeFiles/leveldb.dir/all] Fehler 2
```

#1246 fixes this by requiring C++17 instead of C++11. But you may want to review adding the dependency via `is_standard_layout_v` in the first place.

Simonhancrew:
seems to have been introduced by this [commit](https://github.com/google/leveldb/commit/302786e211d1f2e6fd260261f642d03a91e5922c)

Kosthi:
I also found it.

dingqunfei:
Failed in first step.
cmake and make, then errors.

hq6:
This error is still open. :(

PR #1251, opened 2025-02-13
Fix the issue that prevents compilation in C++11 environment

JalanChao:
issue #1247 
[Commit 302786e ](https://github.com/google/leveldb/commit/302786e211d1f2e6fd260261f642d03a91e5922c) introduced a definition std::is_standard_layout_v that is only supported starting from C++17, which now prevents the LevelDB project from compiling in a C++11 environment.
This modification uses a C++11-compatible implementation, allowing compilation in a C++11 environment without changing the logic. 
The changes are straightforward and do not appear to require performance testing.

### DIFF
--- a/util/env_posix.cc
+++ b/util/env_posix.cc
@@ -874,7 +874,8 @@ class SingletonEnv {
 #endif  // !defined(NDEBUG)
     static_assert(sizeof(env_storage_) >= sizeof(EnvType),
                   "env_storage_ will not fit the Env");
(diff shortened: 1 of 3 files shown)


sihaixianyu:
Awesome, this is a common problem that a bunch of coders has encountered. This PR should be merged in master branch : )

pwnall:

We're adopting [the Google Foundational C++ support policy](https://opensource.google/documentation/policies/cplusplus-support), which sets the minimum supported C++ version to C++17. So, we won't be taking this fix.

PR #1246, opened 2025-01-27
require c++17 instead of c++11

bertsky:
Fixes the following build issue on GCC v9:

```
In file included from leveldb/util/comparator.cc:14:
leveldb/./util/no_destructor.h: In constructor ‘leveldb::NoDestructor<InstanceType>::NoDestructor(ConstructorArgTypes&& ...)’:
leveldb/./util/no_destructor.h:24:24: error: ‘is_standard_layout_v’ is not a member of ‘std’; did you mean ‘is_standard_layout’?
   24 |     static_assert(std::is_standard_layout_v<NoDestructor<InstanceType>>);
      |                        ^~~~~~~~~~~~~~~~~~~~
      |                        is_standard_layout
leveldb/./util/no_destructor.h:24:70: error: expected primary-expression before ‘>’ token
   24 |     static_assert(std::is_standard_layout_v<NoDestructor<InstanceType>>);
      |                                                                      ^~
leveldb/./util/no_destructor.h:24:72: error: expected primary-expression before ‘)’ token
   24 |     static_assert(std::is_standard_layout_v<NoDestructor<InstanceType>>);
      |                                                                        ^
make[2]: *** [CMakeFiles/leveldb.dir/build.make:468: CMakeFiles/leveldb.dir/util/comparator.cc.o] Fehler 1
make[1]: *** [CMakeFiles/Makefile2:183: CMakeFiles/leveldb.dir/all] Fehler 2
```

This is a [C++17 feature](https://en.cppreference.com/w/cpp/types/is_standard_layout).

Note: I won't sign Google CLA.

PR #1260, opened 2025-04-03
Fix compilation errors with C++17 feature

songqing:
Compilation errors with C++17 feature "std::is_standard_layout_v", as the default CMAKE_CXX_STANDARD is C++11, we can change to a compatible implementation.
Related issue https://github.com/google/leveldb/issues/1247 and PR https://github.com/google/leveldb/pull/1246

### DIFF
--- a/util/env_posix.cc
+++ b/util/env_posix.cc
@@ -874,7 +874,7 @@ class SingletonEnv {
 #endif  // !defined(NDEBUG)
     static_assert(sizeof(env_storage_) >= sizeof(EnvType),
                   "env_storage_ will not fit the Env");
-    static_assert(std::is_standard_layout_v<SingletonEnv<EnvType>>);
+    static_assert(std::is_standard_layout<SingletonEnv<EnvType>>::value);
     static_assert(
         offsetof(SingletonEnv<EnvType>, env_storage_) % alignof(EnvType) == 0,
         "env_storage_ does not meet the Env's alignment needs");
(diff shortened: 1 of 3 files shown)


l0rinc:
Or change to C++17 or C++20 instead, like other PRs have done

util/env_posix.cc
```diff
@@ -874,7 +874,7 @@ class SingletonEnv {
 #endif  // !defined(NDEBUG)
     static_assert(sizeof(env_storage_) >= sizeof(EnvType),
                   "env_storage_ will not fit the Env");
-    static_assert(std::is_standard_layout_v<SingletonEnv<EnvType>>);
+    static_assert(std::is_standard_layout<SingletonEnv<EnvType>>::value);
```
l0rinc, line 877: You likely need the message as well to be C++11 compatible
```suggestion
    static_assert(std::is_standard_layout<SingletonEnv<EnvType>>::value,
                    "SingletonEnv<EnvType> is not standard layout");
```

songqing:
> Or change to C++17 or C++20 instead, like other PRs have done

Not all applications using LevelDB support C++17 or higher I think

util/env_posix.cc
l0rinc: See: https://github.com/google/leveldb/pull/1251/files

songqing: ok, haven't seen it before, close this now.

PR #1262, opened 2025-04-04
Do not crash if filesystem can't fsync

l0rinc:
Cherry-pick of https://github.com/bitcoin-core/leveldb-subtree/commit/d42e63d49d9df05b12cd00af4ffc5f2b3edf7e21


felipecrv (changes requested):
This is really bad. Retry with exponential backoff on EINVAL and if after multiple attempts you still get the error, return the error so the caller knows the data hasn't been committed to storage media.

laanwj:
My understanding of the issue was that the reason to make this change was that some file systems fail (CIFS, at the time) instantly when `fsync()` is called on directories. As this is an inherent property of the file system, retrying wouldn't change anything. Returning the error to the caller is what we want to avoid as the software should work with these file systems, even if this means being able to make fewer guarantees.

That said, this was added as an application-specific work-around for a 2018 bug in CIFS it may not be needed anymore, nor necessary to backport.

felipecrv:
> Returning the error to the caller is what we want to avoid as the software should work with these file systems, even if this means being able to make fewer guarantees.

Understandable. The problem is that this solution makes fewer guarantees on every file system, not just CIFS. fsync for CIFS becoming a no-op in the kernel really is the best solution.

PR #1265, opened 2025-04-04
Increase maximum read-only mmap()s used from 1000 to 4096 on 64-bit systems

l0rinc:
Cherry-picking https://github.com/bitcoin-core/leveldb-subtree/commit/92ae82c78f225de84040c51e07fd0b4a61caed99 and revives https://github.com/google/leveldb/pull/613

### DIFF
--- a/util/env_posix.cc
+++ b/util/env_posix.cc
@@ -43,8 +43,8 @@ namespace {
 // Set by EnvPosixTestHelper::SetReadOnlyMMapLimit() and MaxOpenFiles().
 int g_open_read_only_file_limit = -1;
 
-// Up to 1000 mmap regions for 64-bit binaries; none for 32-bit.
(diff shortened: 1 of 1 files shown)


laanwj:
i'm confused, maybe there is value in changing the default, but why do we patch leveldb here when it could be set from client code:
```
// Can be set using EnvPosixTestHelper::SetReadOnlyMMapLimit().
```

l0rinc:
Are you suggesting we revert it in Bitcoin instead?

laanwj:
Ideally yes, looking into it.

laanwj:
Okay this turned out so much worse than i hoped, on the upside, maybe we can just drop this patch: bitcoin/bitcoin#32447

maflcko:
Can this be closed?

l0rinc:
Yes, thanks. Can we also close https://github.com/bitcoin/bitcoin/issues/33351 now?

PR #1300, opened 2026-02-26
Fix CompactRange hang

etiennep-chromium:
This PR ensures that CompactRange doesn't get stuck on waiting on background_work_finished_signal_ during shutdown, because while background_work_finished_signal_ may be signal, CompactMemTable() will never actually run.


pwnall:
Thank you for your contribution! This was merged using the internal contribution process.

PR #1305, opened 2026-03-19
Fix potential integer overflow and unbounded allocation in ReadBlock

Noamismach:
### Description
While reviewing the SSTable parsing logic, I noticed that the varint-decoded `handle.size()` in `ReadBlock` is cast to `size_t` and added to `kBlockTrailerSize` without prior bounds checking or overflow validation.

In scenarios where a corrupted or malformed `.ldb` file is read, an excessively large block size can trigger an integer overflow (`n + kBlockTrailerSize < n`), or force the `new char[]` allocator to request an enormous amount of heap memory, leading to an immediate `std::bad_alloc` abort (process crash).

### Fix
This patch hardens the parser by:
1. Adding an integer overflow guard before the trailer addition.
2. Introducing a safe maximum block size cap (`256MB`) for both compressed and decompressed block payloads.
3. Returning `Status::Corruption` instead of allowing the application to crash, ensuring higher robustness.

All existing tests pass successfully.

### DIFF
--- a/table/format.cc
+++ b/table/format.cc
@@ -4,6 +4,8 @@
 
 #include "table/format.h"
 
+#include <limits>
+
 #include "leveldb/env.h"
 #include "leveldb/options.h"
 #include "port/port.h"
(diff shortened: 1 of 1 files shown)


Noamismach:
Hi team, 
Just a gentle ping on this PR. Let me know if there's anything else you need from my side to approve the workflows or if any adjustments are required for the patch. 

Thanks!

PR #1323, opened 2026-04-27
Fix integer overflow in DecodeEntry bounds check

M0oo0ry:
The bounds check in DecodeEntry() compares `limit - p` against
`*non_shared + *value_length`. Both operands of the addition are
uint32_t, so the sum can wrap around when their combined value
exceeds UINT32_MAX. When that happens the comparison passes
incorrectly and the function returns a pointer that may be past
`limit`, which callers then dereference during block iteration.

This change widens the comparison to uint64_t so the addition cannot
wrap. Behavior on valid blocks is unchanged.

Tested:
- All ctest targets pass on Linux (leveldb_tests, c_test,
  env_posix_test).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -340,6 +340,7 @@ if(LEVELDB_BUILD_TESTS)
         "db/write_batch_test.cc"
         "helpers/memenv/memenv_test.cc"
         "table/filter_block_test.cc"
+	"table/block_test.cc"
         "table/table_test.cc"
         "util/arena_test.cc"
         "util/bloom_test.cc"
--- a/table/block.cc
+++ b/table/block.cc
@@ -68,7 +68,8 @@ static inline const char* DecodeEntry(const char* p, const char* limit,
     if ((p = GetVarint32Ptr(p, limit, value_length)) == nullptr) return nullptr;
   }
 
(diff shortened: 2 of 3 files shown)


xhon-pelushi (approved):
I checked this locally.

- Built PR head `61ad645ff68e0de04abe0aa25e1c014f082fa274` with `cmake -S . -B build-review-1323 -DCMAKE_BUILD_TYPE=Release -DLEVELDB_BUILD_TESTS=ON` and ran `ctest --test-dir build-review-1323 --output-on-failure`: all 3 tests passed.
- Reproduced the overflow against parent `7ee830d02b623e8ffe0b95d59a74db1e58da04c5` with a temporary harness that calls `DecodeEntry()` on an entry header with `shared=0`, `non_shared=UINT32_MAX`, `value_length=1`: the parent accepted it with `remaining=0`, while this PR rejects it.
- Traced the accepted values into `Block::Iter::ParseNextKey()`: `DecodeEntry()` feeds `key_.append(p, non_shared)` and `Slice(p + non_shared, value_length)`, so rejecting the wrapped sum at the decode boundary is the right fix.
- Grepped `table/block.cc` for sibling `non_shared + value_length` style checks and did not find another copy.

Only gap I see is that this PR does not add a dedicated regression test for the wrapped-length entry; the fix itself matches the reproduced failure.

PR #1329, opened 2026-05-24
fix: record new appendable memenv files

### LINKED ISSUE #1328: Bug: NewAppendableFile does not record newly created FileState in file_map_
## Issue Description

I found a potential bug in `helpers/memenv/memenv.cc` in the `InMemoryEnv::NewAppendableFile` method.

## Location
[helpers/memenv/memenv.cc, lines 276-287](https://github.com/google/leveldb/blob/main/helpers/memenv/memenv.cc#L276)

## Problem

When a new appendable file is created (i.e., the file does not exist in `file_map_`), a new `FileState` object is instantiated and referenced. However, **the new FileState is never assigned back to the file_map_**, breaking the association between the filename and the FileState object.

## Current Code

```cpp
Status NewAppendableFile(const std::string& fname,
                         WritableFile** result) override {
  MutexLock lock(&mutex_);
  FileState** sptr = &file_map_[fname];
  FileState* file = *sptr;
  if (file == nullptr) {
    file = new FileState();
    file->Ref();
    // BUG: Missing assignment back to file_map_
  }
  *result = new WritableFileImpl(file);
  return Status::OK();
}

Haihan-Jiang:
## Summary

Fixes #1328.

`InMemoryEnv::NewAppendableFile()` currently creates a new `FileState` for missing files but does not store that pointer back into `file_map_`. The map entry remains null, so later operations that look up the filename can dereference a null `FileState`.

This stores the newly-created `FileState` in the map and adds a regression test that creates a file with `NewAppendableFile()`, appends data, then verifies size and sequential reads through the environment.

## Validation

- `git diff --check`

### DIFF
--- a/helpers/memenv/memenv.cc
+++ b/helpers/memenv/memenv.cc
@@ -281,6 +281,7 @@ class InMemoryEnv : public EnvWrapper {
     if (file == nullptr) {
       file = new FileState();
       file->Ref();
+      *sptr = file;
     }
     *result = new WritableFileImpl(file);
     return Status::OK();
--- a/helpers/memenv/memenv_test.cc
+++ b/helpers/memenv/memenv_test.cc
@@ -91,6 +91,29 @@ TEST_F(MemEnvTest, Basics) {
   ASSERT_LEVELDB_OK(env_->RemoveDir("/dir"));
 }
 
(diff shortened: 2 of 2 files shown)


Haihan-Jiang:
This is ready for review from my side. Could someone take a look when convenient?

Issue #1311, opened 2026-03-22
[Bug] Manifest bad internal key causes DB Open process crash


Sakura-501:
# Manifest bad internal key causes DB Open process crash

## 1. Basic Information

- **Repository**: https://github.com/google/leveldb
- **Affected commit**: `7ee830d02b623e8ffe0b95d59a74db1e58da04c5`
- **Affected version**: leveldb 1.23.0 / HEAD@`7ee830d02b62`
- **Component**: MANIFEST and VersionSet recovery path
- **Primary sink**: `db/version_edit.cc` and `db/version_set.cc`
- **Related path**: `VersionEdit::DecodeFrom()` -> `VersionSet::Builder::Apply()` -> `VersionSet::Recover()` -> `DB::Open()`
- **Vulnerability type**: CWE 191 Integer Underflow leading to process crash

## 2. Vulnerability Summary

`VersionEdit::DecodeFrom()` accepts malformed MANIFEST `smallest` and `largest` values as long as they are non empty. A 1 byte string is therefore accepted as an InternalKey even though a valid InternalKey must contain at least 8 trailing bytes for sequence and type.

Later, when recovery inserts these forged keys into the `VersionSet::Builder` set, `InternalKeyComparator::Compare()` calls `ExtractUserKey()` and performs `internal_key.size() - 8`. For a 1 byte key this underflows, turns into a huge size, and eventually reaches `memcmp`, where AddressSanitizer reports `negative-size-param` and aborts the process.

The result is a crash during `DB::Open()` when a target opens a crafted LevelDB directory.

## 3. Affected Code

```c++
static bool GetInternalKey(Slice* input, InternalKey* dst) {
  Slice str;
  if (GetLengthPrefixedSlice(input, &str)) {
    return dst->DecodeFrom(str);
  } else {
    return false;
  }
}
```

```c++
bool DecodeFrom(const Slice& s) {
  rep_.assign(s.data(), s.size());
  return !rep_.empty();
}

inline Slice ExtractUserKey(const Slice& internal_key) {
  assert(internal_key.size() >= 8);
  return Slice(internal_key.data(), internal_key.size() - 8);
}
```

## 4. Technical Details

[...]

PR #1335, opened 2026-06-11
Reject invalid InternalKeys when decoding MANIFEST edits

uadhran:
## Summary

Fixes a crash during `DB::Open()` when a crafted MANIFEST contains an internal key shorter than 8 bytes (e.g. a 1-byte `smallest`/`largest` in a `kNewFile` record). Previously `InternalKey::DecodeFrom` accepted any non-empty slice, and later `ExtractUserKey()` underflowed `size() - 8` during recovery.

## Changes

- **`db/dbformat.h`**: `InternalKey::DecodeFrom` validates via `ParseInternalKey` (requires >= 8 bytes and valid type).
- **`db/dbformat_test.cc`**: Tests for too-short and invalid-type keys.
- **`db/version_edit_test.cc`**: Regression test that a crafted `kNewFile` edit with a 1-byte internal key returns corruption.

## Testing

```
./leveldb_tests --gtest_filter="FormatTest.InternalKey_*:VersionEditTest.*"  # 6/6 passed
```

Did not re-run the full 212-test suite for this small validation change.

Related: #1311

### DIFF
--- a/db/builder.cc
+++ b/db/builder.cc
@@ -29,10 +29,20 @@ Status BuildTable(const std::string& dbname, Env* env, const Options& options,
     }
 
     TableBuilder* builder = new TableBuilder(options, file);
-    meta->smallest.DecodeFrom(iter->key());
+    if (!meta->smallest.DecodeFrom(iter->key())) {
+      delete builder;
+      delete file;
+      return Status::Corruption("invalid internal key in BuildTable input");
+    }
     Slice key;
+    ParsedInternalKey parsed;
     for (; iter->Valid(); iter->Next()) {
       key = iter->key();
+      if (!ParseInternalKey(key, &parsed)) {
+        delete builder;
+        delete file;
+        return Status::Corruption("invalid internal key in BuildTable input");
+      }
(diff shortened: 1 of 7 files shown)


uadhran:
For context — I noticed #1000 is open on the related #966 (invalid InternalKeys from corrupted files). That PR adds `Comparator::IsInternal()` and guards in `block.cc`, plus a `str.size() < 8` check in `GetInternalKey`.

This PR takes a different path aimed at #1311 specifically: validate at decode time via `ParseInternalKey` in `InternalKey::DecodeFrom` / `GetInternalKey`, so bad MANIFEST keys are rejected during recovery instead of crashing later in `ExtractUserKey`. No public API changes.

Also added defense-in-depth in `ExtractUserKey`, `InternalKeyComparator`, compaction (`db_impl.cc`), and `BuildTable`, plus unit tests for short keys, invalid types, and malformed MANIFEST records.

Happy to consolidate with #1000 if maintainers prefer that approach — just wanted to flag the overlap upfront.

Issue #1310, opened 2026-03-22
[Bug] ReadBlock block size wraparound bypasses truncated block validation and causes out of bounds access


Sakura-501:
# ReadBlock block size wraparound bypasses truncated block validation and causes out of bounds access

## 1. Basic Information
- **Repository**: https://github.com/google/leveldb
- **Affected commit**: `7ee830d02b623e8ffe0b95d59a74db1e58da04c5`
- **Affected version**: leveldb 1.23.0 / HEAD@`7ee830d02b62`
- **Component**: generic SSTable block parsing logic
- **Primary sink**: `table/format.cc` → `ReadBlock()`
- **Related path**: `db/dumpfile.cc` → `Table::Open()` → `ReadBlock()`
- **Vulnerability type**: CWE-190 Integer Overflow / Wraparound, leading to OOB access / DoS

## 2. Vulnerability Summary
`ReadBlock()` narrows attacker-controlled `BlockHandle.size` to `size_t` and then immediately uses it in `n + kBlockTrailerSize`. When `size = UINT64_MAX - 4`, `n + 5` wraps to `0` on a 64-bit build, which means:

1. `new char[n + 5]` allocates using the wrapped tiny length;
2. `file->Read(..., n + 5, ...)` reads using the same wrapped tiny length;
3. the truncated-block check `contents.size() != n + 5` also uses the same wrapped value;
4. the safety check is therefore bypassed in a self-consistent way;
5. the code still performs `switch (data[n])` using the attacker-influenced huge logical size;
6. execution reaches an out-of-bounds pointer computation at `table/format.cc:102`, terminating the process.

This is not limited to test-only code. The bug sits in LevelDB's generic SSTable parsing path, so any program that opens attacker-supplied `.ldb/.sst` files may hit it.

## 3. Affected Code
```c++
size_t n = static_cast<size_t>(handle.size());
char* buf = new char[n + kBlockTrailerSize];
Slice contents;
Status s = file->Read(handle.offset(), n + kBlockTrailerSize, &contents, buf);
...
if (contents.size() != n + kBlockTrailerSize) {
  delete[] buf;
  return Status::Corruption("truncated block read");
}
...
switch (data[n]) {
```

## 4. Technical Details
[...]

PR #1337, opened 2026-06-11
Reject oversized block handles before ReadBlock allocation

uadhran:
## Problem

`ReadBlock()` narrows attacker-controlled `BlockHandle::size()` to `size_t` and computes `n + kBlockTrailerSize` without checking for overflow. When `size` is near `SIZE_MAX` (e.g. `UINT64_MAX - 4`), the addition wraps to a small value. Allocation, the file read, and the truncated-block check all use that wrapped length consistently, so the safety check is bypassed and `data[n]` reads out of bounds.

This affects any code path that opens attacker-supplied SSTables (`Table::Open`, `leveldbutil dump`, etc.).

Related: #1310

## Fix

Validate `handle.size()` as `uint64_t` before narrowing or adding `kBlockTrailerSize`. Return `Corruption("block is too large")` when the size cannot be represented safely.

## Tests

- `ReadBlockTest.RejectsOversizedBlockHandle` — covers the PoC value (`UINT64_MAX - 4`), `UINT64_MAX`, and `SIZE_MAX`
- `ReadBlockTest.TruncatedBlockRead` — block larger than backing file still returns corruption

## Validation (Linux)

```
mkdir -p build && cd build
cmake -DCMAKE_BUILD_TYPE=Release .. && cmake --build .
./leveldb_tests --gtest_filter='ReadBlockTest.*'   # 2/2 passed
./leveldb_tests                                     # 211 passed, 2 skipped
./env_posix_test                                    # not re-run this round
```

`clang-format` was not run (not installed locally).

### DIFF
--- a/table/format.cc
+++ b/table/format.cc
@@ -4,6 +4,8 @@
 
 #include "table/format.h"
 
+#include <limits>
+
 #include "leveldb/env.h"
 #include "leveldb/options.h"
 #include "port/port.h"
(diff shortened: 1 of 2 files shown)


uadhran:
I put together a fix + tests for #1310 — a guard in `ReadBlock()` before `n + kBlockTrailerSize`, plus a couple of unit tests including the `UINT64_MAX - 4` PoC.

After opening this PR, I went back through existing open PRs on the same path and found the same idea is already covered (#1032, #1312, #1308). #1312 is especially close to what I have here, and #1032 has had a similar check since 2022.

I should have checked that more carefully before opening this — my mistake. Closing it now since I don't think another near-duplicate PR helps maintainers in a repo that already barely merges anything.

Keeping #1334, #1335, and #1336 open — those are on different bugs with much less overlap.

Happy to reopen or help review if #1312 / #1032 stall. Thanks.

